PR #22, opened 2016-06-21
Fix generated dependencies to major and minor version

### LINKED ISSUE #11: Dependencies are asterisked
In https://github.com/Microsoft/types-publisher/blob/master/src/lib/package-generator.ts#L93, the dependencies are listed with asterisks. This means any version is compatible. If someone is using an older commit hash and can not update, this will immediately break on them as the dependencies will be bumped to the latest incompatible version.

andy-ms:
Fixes #11


RyanCavanaugh:
:+1:

PR #23, opened 2016-06-23
Support "not needed" packages

andy-ms:
Before merging this we should come up with a list of not-needed packages.


RyanCavanaugh:
:+1:

andy-ms:
Maybe I should empty out the list in `notNeededPackages.json` and merge so we can create the list later?

PR #25, opened 2016-06-24
Add linting

andy-ms:
Quick & dirty, but already caught a few issues.
The unused `const globalAugments` might have been meant for something.


andy-ms:
We might also want to change indent to spaces to be like typescript itself.

mhegazy:
thanks!

mhegazy:
👍

PR #31, opened 2016-06-28
Use promises in parse-definitions

andy-ms:
This appears to be mostly compute-bound so no major savings here.


RyanCavanaugh:
:+1:

PR #35, opened 2016-06-28
Refactor argument parsing up to the script level (never inside functions)

andy-ms:


RyanCavanaugh:
:+1: much better this way

PR #38, opened 2016-06-30
Clean more, and expose as a function

andy-ms:
Also switching from rmdir to rimraf as the package's author [seems](https://github.com/dreamerslab/node.rmdir#description) to be recommending.


RyanCavanaugh:
:+1:

PR #39, opened 2016-06-30
Add script to fetch latest DefinitelyTyped

andy-ms:
I've tested and Azure will allow us to write to a parent directory as well as shell out to `git`.


mhegazy:
:+1:

Issue #40, opened 2016-06-30
This issue tracks the server status


andy-ms:
Server has been up as of **2019-01-29T18:07:45.628+0000**

sandersn:
We have a separate watchdog now, with a badge on Definitely Typed's readme, so this issue is no longer needed. I removed the associated code in #554.

PR #41, opened 2016-06-30
Update an issue on github every time we upload new blobs

andy-ms:
One last thing to do before pulling this is to have a bot create/update the issue. (Otherwise I will be the only one able to do it.) Also to document the GITHUB_ACCESS_TOKEN env var and how to get it.

For now, here's a sample: #40.

### DIFF
--- a/src/lib/blob-uploader.ts
+++ b/src/lib/blob-uploader.ts
@@ -1,5 +1,6 @@
 import assert = require("assert");
 import * as fsp from "fs-promise";
+import moment = require("moment");
 import * as path from "path";
 import Container from "./azure-container";
 import { Logger, ArrayLog } from "./common";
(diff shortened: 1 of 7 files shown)


RyanCavanaugh:
Excellent :+1:

src/lib/blob-uploader.ts
```diff
@@ -3,27 +3,38 @@ import * as fsp from "fs-promise";
...
 import Container from "./azure-container";
 import { Logger, ArrayLog } from "./common";
+import updateIssue from "./issue-updater";
 
 const container = new Container("typespublisher");
 const maxNumberOfOldLogsDirectories = 5;
+const githubAccessToken = process.env.GITHUB_ACCESS_TOKEN;
+
+export default async function uploadBlobsAndUpdateIssue() {
+	const timeStamp = new Date().toISOString();
```
RyanCavanaugh: Let's format this to Seattle timezone

PR #44, opened 2016-07-01
Fix semver bug when there is a prerelease tag

andy-ms:


mhegazy:
👍

thorn0:
Please rerun the publisher ASAP. Now it's impossible to install many packages that have `@types/node@4.0.*` as a dependency.

mhegazy:
i am running it now. it is now processing packages with names starting with c's.  so should be done in the next few hours.

mhegazy:
found another issue, republishing. third time is a charm.

PR #49, opened 2016-07-08
Fix big where version could be parsed across multiple lines

andy-ms:
This was causing `mz` to have a version of `0777`.


mhegazy:
👍

PR #50, opened 2016-07-08
Add `full` task

andy-ms:


src/parse-definitions.ts
```diff
@@ -35,7 +39,7 @@ async function filterPaths(paths: string[]): Promise<{ name: string; path: strin
 	return filterAsyncOrdered(fullPaths, async s => (await fsp.stat(s.path)).isDirectory());
 }
 
-async function main(): Promise<void> {
+export default async function main(): Promise<void> {
```
mhegazy: `export default async function` working.. yaaaay!

mhegazy:
👍

PR #53, opened 2016-07-11
Include subdirectories in search for `.d.ts` files

andy-ms:
Fixes https://github.com/Microsoft/TypeScript/issues/9488
`output\react_router\lib` (among others) will now be generated.

### DIFF
--- a/typings/tsd.d.ts
+++ b/typings/tsd.d.ts
@@ -5,3 +5,4 @@
 /// <reference path="source-map-support/source-map-support.d.ts" />
 /// <reference path="node-fetch/node-fetch.d.ts" />
 /// <reference path="rimraf/rimraf.d.ts" />
+/// <reference path="recursive-readdir/recursive-readdir.d.ts" />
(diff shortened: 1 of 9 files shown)


typings/tsd.d.ts
```diff
@@ -5,3 +5,4 @@
 /// <reference path="source-map-support/source-map-support.d.ts" />
 /// <reference path="node-fetch/node-fetch.d.ts" />
 /// <reference path="rimraf/rimraf.d.ts" />
+/// <reference path="recursive-readdir/recursive-readdir.d.ts" />
```
mhegazy: we should just remove this file and use `@types`.

mhegazy:
👍

mhegazy:
Do you mind doing another publish after this is in.

typings/tsd.d.ts
andy-ms: See #54

dalgard:
Have the types been republished after this fix?

I'm still getting an error with `@types/react-router`.

RyanCavanaugh:
Next republish is scheduled for when the merge from `master` to `types-2.0` finishes, which should be today or tomorrow. After that we'll be publishing on a more consistent rolling basis.

dalgard:
Sounds good – thanks!

PR #58, opened 2016-07-12
Use @types instead of tsd where possible

andy-ms:
Part of #54. We need to move the latest `node-fetch` and `typescript` definitions to the types-2.0 branch of DefinitelyTyped in order to completely get rid of tsd.


RyanCavanaugh:
:+1:

PR #59, opened 2016-07-13
Add webhook

andy-ms:


src/lib/webhook-server.ts
```diff
@@ -0,0 +1,108 @@
...
+	}));
+}
+
+function writeLog(log: ArrayLog): Promise<void> {
+	const { infos, errors } = log.result();
+	assert(!errors.length);
+	return rollingLogs.write(infos);
+}
+
+function listenToGithub(dry: boolean, key: string, onUpdate: (log: ArrayLog, timeStamp: string) => void): Server {
```
RyanCavanaugh: Nit if you feel like it: `key` and `dry` should be in the same order as in `server`

RyanCavanaugh:
:+1:

src/lib/webhook-server.ts
```diff
@@ -0,0 +1,108 @@
...
+				await doOnce(log, timeStamp);
+				working = false;
+			} while (anyUpdatesWhileWorking);
+		}
+	};
+}
+
+function checkSignature(key: string, data: string, actualSignature: string) {
+	const expectedSignature = `sha1=${getDigest()}`;
+	return stringEqualsConstantTime(expectedSignature, actualSignature);
```
RyanCavanaugh: Nit: Probably deserves a comment (e.g. `// prevent timing attacks`) so an overzealous maintainer doesn't remove this

src/lib/webhook-server.ts
```diff
@@ -0,0 +1,108 @@
...
+	assert(!errors.length);
+	return rollingLogs.write(infos);
+}
+
+function listenToGithub(dry: boolean, key: string, onUpdate: (log: ArrayLog, timeStamp: string) => void): Server {
+	return createServer(req => {
+		req.on("data", (data: string) => {
+			const log = new ArrayLog(true);
+			const timeStamp = currentTimeStamp();
+			if (checkSignature(key, data, req.headers["x-hub-signature"])) {
```
RyanCavanaugh: Nit: early return when signature check fails to reduce indentation here

PR #62, opened 2016-07-13
Normalize line endings so content hash is the same on linux and windows

andy-ms:
Since this makes us need to republish everything (current content hashes were computed with `\r\n`), I want to get this in before removing the `alpha` tag and publishing all.


DanielRosenwasser:
LGTM, consider making it `/(\r|\r\n)/g`.

mhegazy:
👍

PR #63, opened 2016-07-13
Error on self-referencing package

andy-ms:
This should eventually be replaced by something in `definition-tester`. @rakatyal


src/lib/definition-parser.ts
```diff
@@ -334,6 +334,10 @@ export async function getTypingInfo(directory: string): Promise<TypingParseFailR
 	const fileContents = await mapAsyncOrdered(declFiles, async d => d + "**" + await readFile(d));
 	const allContent = fileContents.join("||");
 
+	if (libraryName !== "node-schedule" && (referencedLibraries.concat(moduleDependencies).some(s => s === libraryName))) {
```
RyanCavanaugh: What's the deal with `node-schedule` ?

andy-ms: Accidentally left that in while testing.

PR #66, opened 2016-07-14
Allow DefinitelyTyped folders to include a partial `package.json`

andy-ms:


src/lib/package-generator.ts
```diff
@@ -91,44 +91,86 @@ function createMetadataJSON(typing: TypingsData): string {
...
+		typing.moduleDependencies.forEach(addDependency);
+		typing.libraryDependencies.forEach(addDependency);
 	}
-	typing.moduleDependencies.forEach(addDependency);
-	typing.libraryDependencies.forEach(addDependency);
 
-	let version = `${typing.libraryMajorVersion}.${typing.libraryMinorVersion}.${fileVersion}`;
-	if (settings.prereleaseTag) {
-		version = `${version}-${settings.prereleaseTag}`;
+	let version = pkg.version;
```
RyanCavanaugh: We'll need to override the `fileVersion` of a provided version number because NPM doesn't allow publishing a new package with an identical version number

andy-ms: Should we just forbid this field to be specified then?

RyanCavanaugh: Let's block it for now and we can unblock it if we find a reason to need it (I can't think of one, since you can provide the version in the file header)

src/lib/package-generator.ts
```diff
@@ -91,44 +91,86 @@ function createMetadataJSON(typing: TypingsData): string {
...
+		version?: string;
+		description: string;
+	}
+	let pkg: PartialPackageJson = typing.hasPackageJson ?
+		parseJson(await fsp.readFile(pkgPath, { encoding: "utf8" })) :
+		{};
+
+	const ignoredField = Object.keys(pkg).find(field => !["dependencies", "version", "description"].includes(field));
+	if (ignoredField) {
+		throw new Error(`Ignored field in ${pkgPath}: ${ignoredField}`);
```
RyanCavanaugh: This seems to block any of the new logic below from being run?

RyanCavanaugh: Misread, never mind

src/lib/package-generator.ts
```diff
@@ -91,44 +91,85 @@ function createMetadataJSON(typing: TypingsData): string {
...
-	typing.moduleDependencies.forEach(addDependency);
-	typing.libraryDependencies.forEach(addDependency);
 
-	let version = `${typing.libraryMajorVersion}.${typing.libraryMinorVersion}.${fileVersion}`;
-	if (settings.prereleaseTag) {
-		version = `${version}-${settings.prereleaseTag}`;
+	const description = pkg.description || `TypeScript definitions for ${typing.libraryName}`;
+
+	// Don't allow overriding these because they should always be the computed value. E.g. license must always be MIT.
+	const providedFields = ["name", "version", "author", "main", "repository", "scripts", "license", "typings"];
```
RyanCavanaugh: The blacklist code here duplicates the whitelist code in lines 109-111

src/lib/package-generator.ts
```diff
@@ -26,7 +26,7 @@ export async function generatePackage(typing: TypingsData, availableTypes: Types
 	];
 	outputs.push(...typing.files.map(async file => {
```
RyanCavanaugh: Need to rearrange this code a bit because `package.json` will be in the `files` list (so we copy it on line 31) and we'll end up with either the partial or the final file in the output folder depending on who wins the race.

RyanCavanaugh:
:+1: after fixing file copy issue

src/lib/package-generator.ts
andy-ms: Actually, in `definition-parser` we only include the declaration files there.

RyanCavanaugh: Indeed, misread line 136. :+1: then

PR #71, opened 2016-07-14
New algorithm to search for referenced files.

andy-ms:
Now we only include *actually used* files in declFiles.
And we now include files in subdirectories.


src/lib/definition-parser.ts
```diff
@@ -190,49 +178,128 @@ function entryPoint(folderName: string, declFiles: string[], log: string[]): Ent
...
+}
+
+/**
+ * All strings referenced in `import` statements.
+ * Does *not* include <reference> directives.
+ */
+function imports(src: ts.SourceFile): string[] {
+	const out: string[] = [];
+
+	for (const node of src.getChildren()[0].getChildren()) {
```
RyanCavanaugh: `src.getChildren()[0]` could be `src.statements`

RyanCavanaugh:
:+1:

PR #72, opened 2016-07-15
Don't use an issue for logs; just use an `index.html` for that.

andy-ms:
Do use an issue to track webhook server errors.

### DIFF
--- a/src/lib/blob-uploader.ts
+++ b/src/lib/blob-uploader.ts
@@ -2,15 +2,14 @@ import assert = require("assert");
 import * as fsp from "fs-promise";
 import * as path from "path";
 import * as container from "./azure-container";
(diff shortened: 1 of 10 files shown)


src/lib/blob-uploader.ts
```diff
@@ -94,6 +96,25 @@ async function removeOldDirectories(prefix: string, maxDirectories: number, log:
 	await Promise.all(toDelete.map(d => deleteDirectory(prefix + d, log)));
 }
 
-function unique<T>(arr: T[]) {
-	return [...new Set(arr)];
+// Provides links to the latest blobs.
+// These are at: https://typespublisher.blob.core.windows.net/typespublisher/index.html
```
RyanCavanaugh: Please put this in the README.md

RyanCavanaugh:
:+1:

PR #79, opened 2016-07-18
Remove `isSupportedFileKind` since it must always return true.

andy-ms:
Also move DefinitionFileKind to the only file it's used in.


RyanCavanaugh:
:+1:

PR #80, opened 2016-07-19
Don't require an NPM login; use npm-registry-client instead of child_process

andy-ms:


src/settings.d.ts
```diff
@@ -1,4 +1,5 @@
 interface PublishSettings {
+	npmRegistry: string;
```
RyanCavanaugh: Comment would be good on this one

RyanCavanaugh:
:+1:

PR #84, opened 2016-07-21
Add validate script

mhegazy:


andy-ms:
The output of `deleteDirectory` is not used. It could probably just throw an error in that case.

andy-ms:
It looks like every DefinitelyTyped 2.0 project sets noImplicitAny to true, so should be safe to turn that on in these tests.

andy-ms:
This should eventually be integrated into `full.ts`, but right now it's too slow. In the future we should have a version of this that only tests the packages we're about to publish.

andy-ms:
Otherwise, :+1:

andy-ms:
Also, please mention this in the README

RyanCavanaugh:
:+1:

mhegazy:
made it run multiple packages at a time, and accept command line options.

PR #85, opened 2016-07-21
Store versions in an Azure blob

andy-ms:
This also introduces a `changes` file which keeps track of the packages that have actually changed. We can use this to make generation, testing, and publishing faster, but that's a task for another PR.


RyanCavanaugh:
:+1:

PR #86, opened 2016-07-21
GZIP all blobs

andy-ms:
On-disk sizes:

| file | size |
| --- | --- |
| search-index-full | 659kb |
| search-index-min  | 320kb |
| search-index-head | 24kb  |

Uploaded sizes:

| file | size |
| --- | --- |
| search-index-full | 89kb |
| search-index-min  | 78kb |
| search-index-head | 5kb  |

PR #87, opened 2016-07-21
Don't need `search-index-full.json` and `search-index-head.json`, so don't generate them.

andy-ms:
Depends on Microsoft/TypeSearch#13
The files would also have to be manually deleted from Azure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,22 +192,18 @@ This argument may be needed during development,
 
 > `node bin/create-search-index.js`
 
-This script creates several data files useful for offering fast search of types data.
+This script creates `data/search-index-min.json`, which (in the upload step) will be uploaded to Azure and used by [TypeSearch](https://github.com/microsoft/typesearch).
 This step is not necessary for other steps in the process.
 
 ### Arguments to `create-search-index`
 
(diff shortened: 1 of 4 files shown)


RyanCavanaugh:
I'd prefer we keep the `-full` copy around just for debugging purposes. We can get rid of the other two, though

andy-ms:
You mean get rid of just `-head`? `-full` is 11kb bigger than `-min` (see table in #86) so we probably want `-min` to be what TypeSearch accesses. Unless you think the difference isn't worth the complexity.

RyanCavanaugh:
What I meant to say is that I want a file (maybe not a complete file) that has unminified field names in it. We don't have to upload it, just have it be an available artifact for someone to look at.

andy-ms:
Maybe we could have a "--full" flag? Sounds like this is just for a command line users' testing?

RyanCavanaugh:
Sure

andy-ms:
OK, I've redone the entire PR. We now output `-full` as well when the user requests it. Also remembered to document this time.

RyanCavanaugh:
:+1:

PR #93, opened 2016-07-25
Add `make-server-run`

andy-ms:
Also, download gzip from azure, and we don't need special uploading of `versions.json` since it's stored in the `data` folder.


RyanCavanaugh:
:+1:

PR #111, opened 2016-08-15
Create module for logging and log to console more often

andy-ms:


mhegazy:
can we make the loudLogger the default?

andy-ms:
We could change the main functions of `generate-packages`, `parse-definitions`, and `publish-packages` to use loud logging. `package-generator` and `package-publisher` should use quiet logging because these functions are run in parallel with others. How do you mean to make it a default though?

mhegazy:
in stead of requiring the loudLogger from the package, you just require logger, and this is by default is loud. but i see your point about the parallel operations.

mhegazy:
👍

PR #115, opened 2016-08-18
Increase size of IIS stored logs

andy-ms:
This is necessary after #111 - single runs of `full` are taking up many log files because of all the output from the parser and generator.


mhegazy:
:+1: though i do not get why 16 MB specifically for max total size.

andy-ms:
That allows us to store at least 8 logs (if they are all at the maximum individual size).

RyanCavanaugh:
:+1:

PR #116, opened 2016-08-18
Add script to fix all non-gzipped packages

andy-ms:
We always gzip as of #104 but some packages are still non-gzipped. This would republish them.
Running this requires using the `--forceUpdate` option during `calculate-versions`.
This doesn't necessarily need to be merged, since it only needs to be run once.
Should fix DefinitelyTyped/DefinitelyTyped#10508 and DefinitelyTyped/DefinitelyTyped#10506.

### DIFF
--- a/src/calculate-versions.ts
+++ b/src/calculate-versions.ts
@@ -1,14 +1,17 @@
+import * as child_process from "child_process";
+import * as https from "https";
 import * as yargs from "yargs";
-import { existsTypesDataFileSync, readTypings } from "./lib/common";
+import { existsTypesDataFileSync, fullPackageName, readTypings } from "./lib/common";
 import Versions, { Changes, writeChanges } from "./lib/versions";
-import { done } from "./lib/util";
+import { done, streamPromise, unGzip } from "./lib/util";
 
 if (!module.parent) {
 	if (!existsTypesDataFileSync()) {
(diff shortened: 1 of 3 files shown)


mhegazy:
:+1:

andy-ms:
Original version was wrong: the final uploading of `versions.json` would have needlessly updated many packages that we weren't republishing. Fixed to move the check to the calculate-versions step, then we can just publish normally.

andy-ms:
Ran the script. Shall I delete this branch?

RyanCavanaugh:
Not quite sure what's going on here, can you explain in person?

mmckenziedev:
@andy-ms @RyanCavanaugh Did this fix ever get merged? I still experience this issue in my private npm registry proxy.

According to this [NEXUS-10568](https://issues.sonatype.org/browse/NEXUS-10568) they mention that it is being looked at but as far as I can see it has not been resolved.

Can someone please verify and comment?

andy-ms:
The example there uses `@types/colors@0.6.29`. I don't think we can do anything to fix an old version. Do you have any package whose *latest* version doesn't work for you?

mmckenziedev:
@andy-ms This looks more like a problem with my Nexus npm registry proxy. Nothing to do with you guys. Sorry to bother you.

PR #117, opened 2016-08-19
Ensure that the DefinitelyTyped checkout is clean

andy-ms:


RyanCavanaugh:
:+1:. It'd be nice if we could use the Git API directly to ensure this, in case that string changes in the future

RyanCavanaugh:
:+1:

PR #119, opened 2016-08-19
Get rid of `typings`, just use @types/node_fetch and TypeScript's bundled types.

### LINKED ISSUE #54: Use @types instead of tsd internally

andy-ms:
Fixes #54


RyanCavanaugh:
:+1:

PR #121, opened 2016-08-22
Use locally installed versions of npm and tsc for validation

### LINKED ISSUE #120: Don't use `child_process` in `validate`
If we want this to run on azure, we should avoid using `child_process` and instead use APIs directly.

andy-ms:
Fixes #120


mhegazy:
👍

PR #124, opened 2016-08-22
Improve performance of `nAtATime`

andy-ms:
Not that it's a big deal, but I noticed that this function was originally written so that there would be a pause in between batches. This gives about a 25% performance improvement for `npm run index`, which is heavily parallel.
Depends on #123.


RyanCavanaugh:
:+1:

PR #125, opened 2016-08-22
Add script to fix all packages whose README has not been uploaded

andy-ms:
#122 fixes future publishes, but we still have many packages uploaded with no README. This would fix that, just like #116 did for non-gzipped packages. The method is the same: detect the problem by querying npm, and force-update the version of any package that shows the problem.


RyanCavanaugh:
:runner:

andy-ms:
Ran it. Looks like they're all fixed except for the not-needed packages https://www.npmjs.com/package/@types/angular and https://www.npmjs.com/package/@types/moment. Will have to investigate that sometime.

PR #127, opened 2016-08-23
Improvements to readme

andy-ms:
Seems like we should break this up into multiple documents, for example for environment variables or for webhooks.


RyanCavanaugh:
:+1:

PR #129, opened 2016-08-30
Store content hash in published package.json

andy-ms:
What would we ever do if our `versions.json` blob was corrupted or went missing?
This PR would have us start uploading the DefinitelyTyped content hash along with npm packages.

However, NPM already stores a "shasum" along with each package, so if need be, we could calculate the sha of our generated .tgz of of the package and compare it to what's online.
That would be slower, since it requires generating a tgz first rather than just getting the sha of the content on DefinitelyTyped, but this would hopefully be a rare circumstance anyway.

Is this a good idea to merge?

### DIFF
--- a/src/lib/package-generator.ts
+++ b/src/lib/package-generator.ts
@@ -1,7 +1,7 @@
 import { TypesDataFile, TypingsData, NotNeededPackage, fullPackageName, notNeededReadme, settings, getOutputPath } from "./common";
 import { Log, Logger, quietLogger } from "./logging";
 import { readFile, readJson, writeFile } from "./util";
-import Versions from "./versions";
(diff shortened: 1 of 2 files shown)


mhegazy:
looks good to me. @RyanCavanaugh thoughts?

RyanCavanaugh:
:+1: anything that gets rid of versions.json is a good idea IMO

andy-ms:
This PR doesn't get rid of versions.json, but it may be possible if we get the `npm info` of every package and use the "typesPublisherContentHash" (and "lastVersion" would just be the last part of the "version" semver). That may increase the time for "calculate-versions", but do you think it would be worth doing in order to simplify things?

PR #131, opened 2016-08-30
Fix azure-storage version to "1.2.0"

andy-ms:
...because the newest version includes its own typings which override those in declarations.d.ts.


RyanCavanaugh:
:+1:

PR #133, opened 2016-09-01
Fix parsing of version number for libraries with a digit as part of the name

andy-ms:
... (such as "d3-selection")

This should be done before we merge DefinitelyTyped/DefinitelyTyped#10741


mhegazy:
👍

PR #134, opened 2016-09-02
An inferred dependency will always use the latest version

andy-ms:
Fixes Microsoft/TypeScript#10664

What you do think about using `""` instead of `latest`? This would indicate potential compatibility with non-latest versions.
Of course, that might not always be true that it will be compatible with non-latest versions.
But, we of course can't guarantee upon publishing a package that it will forever be compatible with the latest versions of everything else. Using `""` would install the latest by default but allow a user to specify a non-latest version of a dependency in their own `package.json` if that's the version they prefer, and our package would be compatible with that.


mhegazy:
empty is just like "*". so that means older versions as well. not sure if it makes much difference though.

another point to make, should we do this for "global" packages and not for "modules", since modules are safe anyways? does not make a difference?

PR #114, opened 2016-08-16
Fix #10524: Add suport for override entry points in tsconfig.json

mhegazy:
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/10524

flux has two entry points, `flux\index.d.ts` and `flux\utils\index.d.ts`; this adds support for loading these and publishing them. 

the change only impacts 3 packages, flux, angular-mocks, and history. all three are loading additional files that should be loaded.

PR #136, opened 2016-09-08
Include files from tsconfig "files" when inferring all files

andy-ms:
Remake of #114 (so, fixes DefinitelyTyped/DefinitelyTyped#10524)

Note that we no longer error when a package references itself; we just omit the reference from inferred dependencies.

Also, we now assert that DefinitelyTyped folder names are lowercase.

The following packages are affected:
* `aphrodite`: now includes no-important/index.d.ts
* `flux`: now includes utils/index.d.ts
* `history`: now includes lib/DOMUtils.d.ts
* `webpack`: now includes webpack-env.d.ts

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -89,9 +89,9 @@ interface Metadata {
 	projectName: string;
 }
 
-function parseMetadata(entryPointContent: string): Metadata {
+function parseMetadata(mainFileContent: string): Metadata {
 	function regexMatch(rx: RegExp, defaultValue: string): string {
-		const match = rx.exec(entryPointContent);
+		const match = rx.exec(mainFileContent);
 		return match ? match[1] : defaultValue;
 	}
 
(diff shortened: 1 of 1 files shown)


src/lib/definition-parser.ts
```diff
@@ -107,50 +116,58 @@ export async function getTypingInfo(folderName: string): Promise<TypingParseFail
...
 
-	if (packageName !== packageName.toLowerCase()) {
-		log.error(`Package name \`${packageName}\` should be strictly lowercase`);
+	// There is a *single* entry point, containing metadata comments.
+	// But there may be many entryFilenames, which are the starting points of inferring all files to be included.
+	const entryPointResult = await entryPoint(directory, folderName, log.info);
+	if (entryPointResult.kind === "failure") {
+		log.info(entryPointResult.message);
+		log.error(entryPointResult.message);
+		return { kind: "fail", logs: logResult(), rejectionReason: RejectionReason.TooManyFiles };
```
mhegazy: may be we should call this something else other than entryPoint. maybe something like mainFile, packageFile, index.

mhegazy:
👍

PR #139, opened 2016-09-12
Always fetch version info from NPM, rather than the versions.json blob.

andy-ms:
Like `fetch`, we can run this with `nAtATime(100, ...)` and it takes under a minute.
This PR doesn't stop us from uploading the `data` directory along with logs, although we could stop since we no longer download anything from there.
If this is merged into production, it will cause most packages to be republished since we only recently started storing `"typesPublisherContentHash"` on NPM.


RyanCavanaugh:
:+1:

PR #142, opened 2016-09-16
Fail with an error if the definition parser can't find a main file

andy-ms:
Depends on DefinitelyTyped/DefinitelyTyped#11280 and DefinitelyTyped/DefinitelyTyped#11282


mhegazy:
👍

andy-ms:
Updated it so it fails if the main file is not `index.d.ts`. Depends on DefinitelyTyped/DefinitelyTyped#12834.

PR #143, opened 2016-09-16
Forbid "../" in references rather than just ignoring it

### LINKED ISSUE #69: Warn for relative references that reach outside of their directory
Packages in DefinitelyTyped should use global references for this, rather than `../some-other-library`.

andy-ms:
Fixes #69

Depends on DefinitelyTyped/DefinitelyTyped#11271


mhegazy:
This needs to be in the definition tester. this will just block publishing the package, which is not bad, but it is a bit too late.

RyanCavanaugh:
:+1: we should start on the DT / T-P refactoring soon

andy-ms:
Now that types-publisher *is* the tester, this should be safe to do; parsing is part of testing.

PR #144, opened 2016-09-20
Reduce parallelism to avoid overloading npm.

andy-ms:
We were getting `FetchError` in production (both `ETIMEDOUT` and `ECONNRESET`).
Also, allow "scripts" to appear in `package.json` and be ignored.


RyanCavanaugh:
:+1: seems fine to hardcode it TBH

PR #145, opened 2016-09-22
Add script to check that NPM website has correct versions

andy-ms:


andy-ms:
An issue was filed at: npm/www#34
Which is now tracked at: npm/registry#42

It's not a types-publisher specific issue, so we probably won't need this script again.

PR #149, opened 2016-09-29
Allow a partial package.json to specify peerDependencies

andy-ms:
We were discussing using peerDepdendencies as a better solution to DefinitelyTyped/DefinitelyTyped#11547.


RyanCavanaugh:
:+1:

PR #150, opened 2016-09-29
Update TypeScript version so we can parse the type `0 | 1`

andy-ms:
Used [npm-check-updates](https://github.com/tjunnone/npm-check-updates).


RyanCavanaugh:
:+1:

PR #154, opened 2016-10-03
Detect imports inside of module declarations

andy-ms:
This detects the nested import in`mongoose-seeder` (although we should convert that to an external module anyway).


src/lib/definition-parser.ts
```diff
@@ -289,32 +289,45 @@ function referencedFiles(src: ts.SourceFile, subDirectory: string): string[] {
...
+						out.push(parseRequire(decl.moduleReference.getText()));
+					}
+					break;
 				}
-				break;
-			}
 
-			default:
+				case ts.SyntaxKind.ModuleDeclaration: {
+					let decl = node as ts.ModuleDeclaration;
```
mhegazy: you could also add a check that the module name is string to avoid walking namespaces.

PR #156, opened 2016-10-03
Use builtin `crypto.timingSafeEqual`

andy-ms:
This [function](https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b) is back in core node.js, so we don't need to use an external library any more.

Just to be sure, tested with the expected header, a header with the wrong length, and a header with the right length but wrong content.


RyanCavanaugh:
:+1:

PR #157, opened 2016-10-04
Use Azure KeyVault instead of storing passwords directly in environment variables

### LINKED ISSUE #147: Save NPM login token
We should cache the result of the `logIn` function in `npm-client.ts`.
We could do this in a local file, but this could be a security issue.
Maybe we should delay this until we are using KeyVault.

andy-ms:
The keyvault can be viewed under `types-publisher-keys` on [azure](https://ms.portal.azure.com).
If this PR is merged, I will go to our types-publisher azure website and delete all current environment variables, then add `TYPES_PUBLISHER_CLIENT_ID` and `TYPES_PUBLISHER_CLIENT_SECRET`.

**EDIT**: Now fixes #147.


src/lib/npm-client.ts
```diff
@@ -77,6 +75,8 @@ export default class NpmClient {
 }
 
 async function logIn(client: RegClient, password: string): Promise<RegClient.Credentials> {
+	console.log(password);
```
RyanCavanaugh: Remove

src/lib/secrets.ts
```diff
@@ -0,0 +1,79 @@
...
+	GITHUB_ACCESS_TOKEN,
+	/**
+	 * This is used to ensure that only GitHub cand send messages to our server.
+	 * This should match the secret value set on GitHub: https://github.com/DefinitelyTyped/DefinitelyTyped/settings/hooks
+	 * The Payload URL should be the URL of the Azure service.
+	 * The webhook ignores the `sourceRepository` setting and can be triggered by *anything* with the secret, so make sure only DefinitelyTyped has the secret.
+	 */
+	GITHUB_SECRET,
+	/** Password for settings.npmUsername */
+	NPM_PASSWORD
```
RyanCavanaugh: Should we be using an NPM token instead?

RyanCavanaugh:
:+1:

andy-ms:
Successfully ran with the online publsher. :icecream:

PR #190, opened 2016-11-15
Add ability to lint DefinitelyTyped

andy-ms:
If merged, I will also add `tslint.json` creation to the new package generator (DefinitelyTyped/DefinitelyTyped#12521) and make another PR describing this in the README.
Also still planning to do a rule for `<reference path="../"/>`, just ran out of time today.


src/tslint/noSingleDeclareModuleRule.ts
```diff
@@ -0,0 +1,53 @@
...
+	static metadata: Lint.IRuleMetadata = {
+		ruleName: "no-single-declare-module",
+		description: "Don't use an ambient module declaration if you can use an external module file.",
+		rationale: "Cuts down on nesting",
+		options: {},
+		type: "style"
+	};
+
+	static FAILURE_STRING = "File has only 1 module declaration — write it as an external module.\n" +
+		"If augmenting a single module, use `// tslint:disable-next-line:no-single-declare-module`.";
```
mhegazy: I do not think this is needed. just make sure the file is not a module before you call applyWithWalker.

PR #194, opened 2016-11-21
Write new header parser with strict mode for linting

andy-ms:
We will now lint the header by default.
Related: DefinitelyTyped/DefinitelyTyped#12829, DefinitelyTyped/DefinitelyTyped#12833
Inspired by https://github.com/DefinitelyTyped/definition-header


RyanCavanaugh:
:+1:

PR #203, opened 2016-11-29
Catch `interface I extends Function { (): void; }` as a functional interface, and provide a better error message

andy-ms:


RyanCavanaugh:
:+1:

PR #207, opened 2016-12-05
Move package.json error reporting to tester

andy-ms:
See DefinitelyTyped/DefinitelyTyped#13113


blakeembrey:
Quick comments:

1. Erroring on these fields won't work once redirects are setup - all of @types uses it for testing
2. The check (https://github.com/Microsoft/types-publisher/pull/207/files#diff-3cd3598c6c892a75812c78c325162a4eR176) for `scripts` is buggy - it won't error on bad fields when it encounters `scripts` _first_ (e.g. `['scripts', 'devDependencies']` which allows for a workaround of this check).

andy-ms:
For 2, see #208. For 1, we'll deal with when we update for redirects.

Issue #205, opened 2016-12-01
Certain published npm module version numbers reset


pe8ter:
Certain published typing module version numbers have reset back to 0.0.0 and are counting up from there. For example [@types/node](https://www.npmjs.com/package/@types/node) is now at 0.0.2 where not too long ago it was at 6.0.51.

I'd expect its version to still be in the 6.x range.

andy-ms:
Fixed by https://github.com/Microsoft/types-publisher/pull/209

IdeaHunter:
@andy-ms Im still experience the same problem with [@types/sharepoint](https://www.npmjs.com/package/@types/sharepoint): the latest commit merged 2 days ago DefinitelyTyped/DefinitelyTyped#13335, npm show current version as `0.0.0` - but the latest one according to semver should be `2013.0.31`.

PR #209, opened 2016-12-06
Allow version in header to end in 'x'

andy-ms:
Fixes DefinitelyTyped/DefinitelyTyped#13032
I did a diff with the previous `definitions.json` and it looks good. Example:
```diff
1930,1932c1930,1932
<         "libraryMajorVersion": 0,
<         "libraryMinorVersion": 0,
<         "libraryName": "angular-strap v2.2.x",
---
>         "libraryMajorVersion": 2,
>         "libraryMinorVersion": 2,
>         "libraryName": "angular-strap",
```

### DIFF
--- a/src/lib/header.ts
+++ b/src/lib/header.ts
@@ -123,7 +123,8 @@ function parseLabel(strict: boolean): pm.Parser<Label> {
 		// Parse in reverse. Once we've stripped off the version, the rest is the libary name.
 		const reversed = reverse(tilNewline);
 
-		const rgx = /((\d+)\.(\d+)(\.\d+)?(v)? )?(.+)/;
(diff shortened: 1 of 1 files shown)


pe8ter:
I know I don't have any insight into when the new version of the tool that contains this fix will be released, but when it does will you run automatically to fix the errant versions that got published to npm or will we have to 'goose' the tool by submitting updated typings to DefinitelyTyped to force the tool to publish a new version with correct version numbers?

For context, I'm referring to #205, which this PR purportedly fixes.

andy-ms:
#218 should fix that too, since we'll look for another release with the same major/minor and see that it's out-of-date.

PR #212, opened 2016-12-08
Add lint rule: Only allow `void` as return type

andy-ms:
Would this be a good rule? Spotted a stray `void` at DefinitelyTyped/DefinitelyTyped#13191.

### DIFF
--- a/src/tslint/voidReturnRule.ts
+++ b/src/tslint/voidReturnRule.ts
@@ -1,0 +1,51 @@
+import * as Lint from "tslint";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.AbstractRule {
+	static metadata: Lint.IRuleMetadata = {
+		ruleName: "void-return",
+		description: "`void` may only be used as a return type.",
+		rationale: "style",
+		optionsDescription: "Not configurable.",
+		options: null,
+		type: "style",
+		typescriptOnly: true,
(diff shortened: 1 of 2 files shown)


Bomret:
But how would one then write a signature for the following javascript function from [content-disposition](https://github.com/jshttp/content-disposition)?

```
function contentDisposition(filename, options);
```

Quote from the website:
> The filename is optional and if no file name is desired, but you want to specify options, set filename to undefined.

So `filename`may be set to `undefined` but then `options` has to be set. Making `filename` an optional parameter and `options` not would not be allowed by tsc.

Bomret:
Ok, just by declaring it like

```
function contentDisposition(filename: undefined: options: Options);
```

src/tslint/voidReturnRule.ts
```diff
@@ -0,0 +1,51 @@
...
+            const parent = node.parent!;
+            if (!(isSignatureDeclaration(parent) && parent.type === node)) {
+                this.fail(node, Rule.FAILURE_STRING);
+            }
+        }
+        super.visitNode(node);
+    }
+
+    private fail(node: ts.Node, message: string) {
+		this.addFailure(this.createFailure(node.getStart(), node.getWidth(), message));
```
RyanCavanaugh: Tabs

PR #218, opened 2016-12-09
Look for previously-published package versions with the same major/minor version when detecting "old" version.

andy-ms:
This means that if we publish node `6.0.51` and then node `0.0.0` due to bad version parsing, the next published will still be `6.0.52`.
Should do this before DefinitelyTyped/DefinitelyTyped#13239.

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -143,3 +143,7 @@ export function min<T>(inputs: T[], lessThan: (a: T, b: T) => boolean): T | unde
 	}
 	return min;
 }
+export function max<T>(inputs: T[], greaterThan: (a: T, b: T) => boolean): T | undefined {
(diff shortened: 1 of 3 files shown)


src/util/util.ts
```diff
@@ -143,3 +143,7 @@ export function min<T>(inputs: T[], lessThan: (a: T, b: T) => boolean): T | unde
 	}
 	return min;
 }
+export function max<T>(inputs: T[], greaterThan: (a: T, b: T) => boolean): T | undefined {
+	// Logic is the same either way.
+	return min(inputs, greaterThan);
```
RyanCavanaugh: Nit: maybe this function should just be called `best` and takes an `isBetter` argument. Just a suggestion.

PR #219, opened 2016-12-12
Add lint rule for type parameters that don't appear in any parameter type

andy-ms:
This warns for `getMeAT<T>(): T`.


RyanCavanaugh:
Not sure we'll be able to turn this on (lots of people like this despite my personal opinion being different) but the code looks fine

mhegazy:
it can be helpful in certain cases as a alternative for cast., e.g. `Promise.reject()`

andy-ms:
In that particular case, you could return `Promise<never>`. But I see the point about `foo<T>()` being nicer-looking than `foo() as T`.

PR #229, opened 2016-12-15
Allow packages to target typescript@next

### LINKED ISSUE #214: Need a long term solution to supporting users on bleeding edge versions of typescript
It makes sense to not want to break existing users with a "^" dependency on something like `@types/react`, but it hinders the very users and organizations that help the TS ecosystem. The introduction of `Partial<>` and co in particular is a big game changer, and many users would benefit from updates to react and lodash in particular.

The current policy has been stated as waiting 30 days before allowing new features into DefinitelyTyped to avoid breaking people. While I disagree with the policy, I think it we as a community can do better to support those on the bleeding edge.

Considering the situation, I'd like to propose a few options.

**Option A**, do nothing. There are work arounds if you want to start moving fast, such as hacking `paths` and `typeRoots` in your tsconfig. I have documented one such solution here: https://medium.com/@ericlanderson/using-custom-typescript-definitions-with-ts-2-x-3121db84015d#.tvqw3rmth

While this works, its a hassle.

**Option B**: Someone (maybe me) forks DefinitelyTyped and allows changes for the latest and greatest right away. I could publish this as something like "types-edge" in npm and provide documentation for how to configure your tsconfig to allow for it to work. 

While this works, it half forks the community, and creates a burden on myself to keep up to date with DefinitelyTyped. It also hinders discoverability. I think this hurts the ecosystem, but its better than everyone manually forking type definitions.

**Option C**: DefinitelyTyped and Microsoft support the development of these edge versions on a branch within DefinitelyTyped (lets call it "edge" or even use "master"). Microsoft could then register the org @types-edge and publish the edge branch. The DefinitelyTyped and Microsoft people could then as a group keep merging types-2.0 (which should really be called "stable" in this world) into edge, while allowing the community to grow and move quickly. If desired, there could also be a "nightly" branch and a corresponding @types-nightly (I would have personally liked this as the second `Partial<>` and `keyof` got merged, I started using it immediately). This would also mean the types are already being updated and ready to go when the next TS release happens.

There could also be a release of typescript 2.1.1 pretty quickly that adds "./node_modules/@types-edge" to the default list of for the `paths` and the `typeRoots` in tsconfig. 

I think this is the best short-term solution.

**Option D**: Add macros to TS. (Yeah this is dirty). But the macros could allow things like `#if typescriptVersion >= 2.1.0` and allow additional definitions to be placed into the type files. This would still require the 30 day grace period before it could be introduced (maybe in TS 2.2?) but would future proof this problem and allow just a single npm org for publishing types. Its also not backwards compatible, but it would be the easiest to maintain in DefinitelyTyped.

**Option E**: Add additional meta-data support to package.json to declare types with a minimum version. This could look something like:
```json
{
  "...": "...",
  "types2": {
    ">=2.1.0": "index-2.1.0.d.ts"
  }
}
```

Newer versions of typescript can look at this instead of `types` and choose the highest version that maps.

This would be a PITA to support, IMHO. You could do it with branches in DT, but now we have to support forward merging changes to types-2.0 to types-2.1 and then types-2.2, etc. Or you have to keep the different versions in the same branch and remember to change them all when adding new functionality.

Another way to do this, which I have no idea if it would work given future changes to TS are unknown, is write scripts that convert the latest typings into older versions. For example, pretty much every typing that would want `Partial<T>` could just be mutated at the time the publisher runs to replace it with `T`. `Pick` is probably `any` in the old world. `ReadonlyArray` support wouldn't have been broken cause it could just be `Array`. This would require custom scripts for each new release but it might not be awful.

There are probably other ways to do it, but these are the ones that came to me today. I would really love to hear the opinions of a bunch of people: @andy-ms, @ahejlsberg, @pspeter3 @vsiao @johnnyreilly, @jkillian, and more.

Thanks

andy-ms:
Fixes #214

* Add a `typeScriptVersion` property for each definition. This may be overridden by an extra line added to the header.
* If we deem it a prerelease version, publish it as MAJOR.MINOR.0-next.PATCH
* Install multiple typescript versions, so we use the appropriate one for tests.
* Add a script for tagging all packages when we make a change. Normally, just tag a package when publishing it.

### DIFF
--- a/src/npmTags.ts
+++ b/src/npmTags.ts
@@ -1,6 +1,6 @@
 import * as yargs from "yargs";
 
-import { AnyPackage, Options, TypeScriptVersion, existsTypesDataFileSync, fullPackageName, readAllPackages } from "./lib/common";
+import { AnyPackage, TypeScriptVersion, existsTypesDataFileSync, fullPackageName, readTypings } from "./lib/common";
 import NpmClient from "./lib/npm-client";
 import Versions, { versionString } from "./lib/versions";
 
@@ -16,7 +16,7 @@ if (!module.parent) {
 	}
 	else {
 		const dry = !!yargs.argv.dry;
-		done(tagAll(Options.defaults, dry));
+		done(tagAll(dry));
 	}
 }
 
@@ -25,19 +25,16 @@ if (!module.parent) {
  * This shouldn't normally need to run, since we run `tagSingle` whenever we publish a package.
  * But this should be run if the way we calculate tags changes (e.g. when a new release is allowed to be tagged "latest").
  */
-async function tagAll(options: Options, dry: boolean) {
+async function tagAll(dry: boolean) {
 	const versions = await Versions.load();
 	const client = await NpmClient.create();
 
-	const { typings, notNeeded } = await readAllPackages(options);
-
-	for (const t of typings) {
+	for (const t of await readTypings()) {
 		const version = versionString(versions.versionInfo(t).version);
 		await addNpmTagsForPackage(t, version, client, console.log, dry);
 	}
 
-	//TODO: Do it for notNeeded too!
-	notNeeded;
+	// Don't tag notNeeded packages
 }
 
 export async function addNpmTagsForPackage(pkg: AnyPackage, version: string, client: NpmClient, log: Logger, dry: boolean): Promise<void> {
--- a/src/npmTags.ts
+++ b/src/npmTags.ts
@@ -1,0 +1,51 @@
+import * as yargs from "yargs";
+
+import { AnyPackage, Options, TypeScriptVersion, existsTypesDataFileSync, fullPackageName, readAllPackages } from "./lib/common";
+import NpmClient from "./lib/npm-client";
+import Versions, { versionString } from "./lib/versions";
+
+import { Logger } from "./util/logging";
+import { done } from "./util/util";
+
+if (!module.parent) {
+	if (!existsTypesDataFileSync()) {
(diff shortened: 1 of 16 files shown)


blakeembrey:
I mentioned this in the Gitter chat for DefinitelyTyped-Future, but this breaks semver. Most importantly, it also makes it difficult to specify a dependency or peer dependency on the regular version _and_ have it also work for the `@next` version and I imagine many people will miss that nuance.

src/npmTags.ts
```diff
@@ -0,0 +1,51 @@
...
+	const client = await NpmClient.create();
+
+	const { typings, notNeeded } = await readAllPackages(options);
+
+	for (const t of typings) {
+		const version = versionString(versions.versionInfo(t).version);
+		await addNpmTagsForPackage(t, version, client, console.log, dry);
+	}
+
+	//TODO: Do it for notNeeded too!
```
RyanCavanaugh: ?

andy-ms: Actually, I don't think we need to tag the notNeeded packages. If someone installs `foo@2.0`, we don't want them depending on a newer version of `foo` which requires TS 2.1.

andy-ms:
@blakeembrey That's too bad, but not sure what we can do about that -- such is the nature of prerelease versions. At least "normal" users can just wait a month for the prerelease to get published as a regular release.

blakeembrey:
@andy-ms Well, the simple work-around would be to patch increment every release because patches don't mean anything already, then just move the `latest` tag around for NPM.

PR #255, opened 2016-12-30
Update unified-signatures to be like the one for tslint

andy-ms:
Since palantir/tslint#1944 made some improvements, adding them here.
We no longer consider `static foo()` and `foo()` to be overloads. This should fix the lint error in DefinitelyTyped/DefinitelyTyped#13628.

### DIFF
--- a/src/tslint/unifiedSignaturesRule.ts
+++ b/src/tslint/unifiedSignaturesRule.ts
@@ -1,18 +1,40 @@
-import * as Lint from "tslint";
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import * as ts from "typescript";
 
+import * as Lint from "tslint";
+
 export class Rule extends Lint.Rules.AbstractRule {
+	/* tslint:disable:object-literal-sort-keys */
 	static metadata: Lint.IRuleMetadata = {
-		ruleName: "array-type-style",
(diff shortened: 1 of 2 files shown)


src/tslint/unifiedSignaturesRule.ts
```diff
@@ -24,241 +46,284 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-		const typeParametersSet = new Set((typeParameters || []).map(t => t.getText()));
-		for (const overloads of collectOverloads(signatures).values()) {
-			for (const [a, b] of pairs(overloads)) {
-				this.compareSignatures(a, b, type => typeParametersSet.has(type));
-			}
+	private checkMembers(members: Array<ts.TypeElement | ts.ClassElement>, typeParameters?: ts.TypeParameterDeclaration[]) {
+		this.checkOverloads(members, getOverloadName, typeParameters);
+		function getOverloadName(member: ts.TypeElement | ts.ClassElement) {
+			const key = getOverloadKey(member);
+			return key === undefined ? undefined : { signature: member as any as ts.SignatureDeclaration, key };
```
RyanCavanaugh: Why do we need two casts here?

Unrelated: shorter version is `key && { signature ...`

andy-ms: 1) `Type 'TypeElement | ClassElement' cannot be converted to type 'SignatureDeclaration'` because the compiler doesn't know that `getOverloadKey` checks this.
2) Key could be `""` (It shouldn't be, but compiler complains)

src/tslint/unifiedSignaturesRule.ts
```diff
@@ -24,241 +46,284 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+/**
+ * Detect `a(): void` and `a(x: number): void`.
+ * Returns the parameter declaration (`x: number` in this example) that should be optional.
+ */
+function signaturesDifferByOptionalParameter(types1: ts.ParameterDeclaration[], types2: ts.ParameterDeclaration[],
+	): ts.ParameterDeclaration | undefined {
+	const minLength = Math.min(types1.length, types2.length);
+	// Lengths must differ by 1.
+	if (Math.abs(types1.length - types2.length) !== 1) {
```
RyanCavanaugh: Isn't this equally incorrect for adding 2 optional parameters?

andy-ms: You're right. We should catch `f(): void` and `f(x?: number, y?: number): void`. Will fix later.

andy-ms:
No longer needed as we now use the one from tslint. #271

PR #256, opened 2016-12-30
Parse a dependency for "mz" instead of "mz/fs"

### LINKED ISSUE #239: definition-parser: Handle dependencies with "/"
`@types/fs-promise` depends on `"mz/fs"`, so it should include `"@types/mz"` as a dependency.

andy-ms:
Fixes #239

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -248,8 +248,9 @@ async function getModuleInfo(directory: string, folderName: string, allEntryFile
 
 		for (const ref of imports(src)) {
 			if (!ref.startsWith(".")) {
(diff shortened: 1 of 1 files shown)


src/lib/definition-parser.ts
```diff
@@ -365,6 +366,12 @@ async function getModuleInfo(directory: string, folderName: string, allEntryFile
 	}
 }
 
+/** Given "foo/bar/baz", return "foo". */
```
mhegazy: that is true as long as scoped packages are not supported. `@foo/bar/baz`'s root is `@foo/bar`.

PR #259, opened 2016-12-30
Add lint rule for empty namespaces

andy-ms:


mhegazy:
This is needed where you want to enable `import *` on a module with only `export = <function>`. so what is the alternative here?

andy-ms:
Maybe we should do something about Microsoft/TypeScript#5073 instead.

mhegazy:
that would be more economical.

PR #269, opened 2017-01-10
Add additional check for path mapping: an older version must map to itself

andy-ms:
Errors found are fixed in DefinitelyTyped/DefinitelyTyped#13902.

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -1,9 +1,10 @@
 import * as fsp from "fs-promise";
 import * as path from "path";
+import * as ts from "typescript";
 
 import { readFile as readFileText } from "../util/io";
 import { Log, moveLogs, quietLogger } from "../util/logging";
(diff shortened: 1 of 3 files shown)


src/lib/definition-parser.ts
```diff
@@ -154,34 +156,41 @@ async function entryFilesFromTsConfig(packageName: string, directory: string): P
...
-	return res;
+	return makeObject(dependencies, dependency => {
+		const path = paths && paths[dependency];
+		return path === undefined ? "*" : parseDependencyVersionFromPath(packageName, dependency, path);
+	});
 }
 
 // e.g. parseDependencyVersionFromPath("../../foo/v0", "foo") should return "0"
-function parseDependencyVersionFromPath(packageName: string, dependencyName: string, dependencyPath: string): number {
-	let short = dependencyPath;
```
andy-ms: We don't need parent directories because `baseUrl` should be set.

PR #292, opened 2017-01-22
Allow packages to opt-in to typings-checker instead of tsc

danvk:
Just throwing this out there to get some feedback. See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/1572#issuecomment-272571498

For packages that opt-in, this would let them run their tests through `typings-checker` instead of `tsc`. `typings-checker` lets allows errors to be marked as expected and lets tests make assertions about the types of symbols. This helps prevent unintended `any` or `{}` types from creeping in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "tar": "^2.2.1",
     "tslint": "^4.2.0",
     "typescript": "next",
+    "typings-checker": "^1.1.1",
     "yargs": "^6.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


danvk:
See also https://github.com/danvk/typings-checker/issues/18 — it might be easier to set up `typings-checker` as something that runs in addition to `tsc` rather than instead of it (just like `tslint`).

andy-ms:
I'm working on a new tester at https://github.com/andy-ms/dtslint (eventually this will be moved to https://github.com/Microsoft/dtslint). This will allow us to have just one tool instead of `types-publisher` + `tsc` + `tslint` + `typings-checker`. It includes a conversion of `typings-checker` into a lint rule at https://github.com/andy-ms/dtslint/blob/master/src/rules/expectRule.ts . Would this be OK?

I also removed the ability for `$ExpectError` to specify an error string, because we want to be able to change error messages without breaking any tests. You had already put that functionality under a flag.

danvk:
@andy-ms cool! Glad to hear the TS team is improving tooling around testing typings files.

Feel free to use whatever code you like, just be sure to mention me as a contributor.

andy-ms:
Great! Closing in favor of #312.

PR #294, opened 2017-01-31
Store TypeScript version tags in typings registry

andy-ms:


mhegazy:
We need to add some throttling on the publishing of types-registry. possibly publish it once a day at most.

PR #296, opened 2017-02-13
fix(diff): omit removed files

acrazing:
This will avoid throw error when remove a version's typings definitions.


andy-ms:
This should be handled by `tryGetTypingsData` already, but it had a bug for older versions. See #297.

acrazing:
I think the removed files should not be checked. So it should be omitted in `gitDiff` or `gitChanges` method.

andy-ms:
The removed files should not be checked now that #297 is in.

PR #321, opened 2017-03-15
Fix handling of test dependencies

andy-ms:
Continuation of #315
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14557#issuecomment-286829607

* Don't use `.imports` because that's not set. Use our code for walking the tree to find imports.
* Use `.allDependencyTypes` instead of `.dependencyTypes` in check-parse-results and get-affected-packages

### DIFF
--- a/src/tester/get-affected-packages.ts
+++ b/src/tester/get-affected-packages.ts
@@ -23,11 +23,6 @@ export default async function getAffectedPackages(allPackages: AllPackages, log:
 	return collectDependers(changedPackages, dependedOn);
 }
 
-/** Every package name in the original list, plus their dependencies (incl. dependencies' dependencies). */
-export function allDependencies(allPackages: AllPackages, packages: TypingsData[]): TypingsData[] {
-	return sortPackages(transitiveClosure(packages, pkg => allPackages.dependencyTypings(pkg)));
-}
-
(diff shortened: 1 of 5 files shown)


src/tester/test-runner.ts
```diff
@@ -55,7 +55,7 @@ export default async function main(options: Options, nProcesses?: number, regexp
 
 	console.log("Installing dependencies...");
 
-	await nAtATime(nProcesses, allDependencies(allPackages, typings), async pkg => {
+	await nAtATime(nProcesses, typings, async pkg => {
```
andy-ms: We got the dependencies as part of `getAffectedPackages`. So don't need to get them again.

src/tester/get-affected-packages.ts
```diff
@@ -124,10 +119,10 @@ async function gitDiff(log: Logger, options: Options): Promise<string[]> {
 
 	// `git diff foo...bar` gets all changes from X to `bar` where X is the common ancestor of `foo` and `bar`.
 	// Source: https://git-scm.com/docs/git-diff
-	let diff = (await run(`git diff ${sourceBranch}...HEAD --name-only`)).trim();
```
andy-ms: Don't need the `...HEAD`; this enables running the tester on non-committed changes.

PR #324, opened 2017-03-28
fix(package generator): incorrect exported path

alan-agius4:
At the moment the path is "Files were exported from `https://www.github.com/DefinitelyTyped/DefinitelyTyped/tree/master/stylelint`" however the path should be `https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/stylelint`

Note: missing `/types`


andy-ms:
Thanks!

PR #325, opened 2017-03-28
ignore .DS_Store files

scttcper:
finder randomly creates these files inside the folders and prevents tests from being run on osx


andy-ms:
Should be fixed by #312: you won't need to run `types-publisher` yourself, just `dtslint`.

PR #344, opened 2017-05-09
Fix too rigid dtslint path

flying-sheep:
When installing `types-publisher` with e.g. yarn, or flattening the module tree, the hardcoded path will get invalid.

This uses node’s import machinery to find the path.

As dtslint’s `main` field in package.json points to `bin/index.js`, this will result in the right path.


andy-ms:
Thanks!

PR #356, opened 2017-06-23
Allow path mapping for scoped packages

andy-ms:
The problem is that in DefinitelyTyped we don't use `node` module resolution, we use a baseUrl and classic resolution. So scoped package names are not resolved. This allows us to use path mapping for those.

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -223,9 +223,9 @@ async function calculateDependencies(packageName: string, tsconfig: TsConfig, de
 		const pathMapping = pathMappingList[0];
 
 		// Path mapping may be for "@foo/bar" -> "foo__bar". Based on `getPackageNameFromAtTypesDirectory` in TypeScript.
-		const mangledScopeedPackageSeparator = "__";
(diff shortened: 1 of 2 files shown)


src/lib/definition-parser.ts
```diff
@@ -216,7 +216,22 @@ async function calculateDependencies(packageName: string, tsconfig: TsConfig, de
...
 
-		const pathMapping = paths[dependencyName];
+		const pathMappingList = paths[dependencyName];
+		if (pathMappingList.length !== 1) {
+			throw new Error(`In ${packageName}: Path mapping for ${dependencyName} may only have 1 entry.`);
+		}
+		const pathMapping = pathMappingList[0];
+
+		// Path mapping may be for "@foo/bar" -> "foo__bar". Based on `getPackageNameFromAtTypesDirectory` in TypeScript.
+		const mangledScopeedPackageSeparator = "__";
```
joscha: There is a typo here, `mangledScopeedPackageSeparator` -> `ee`.

joscha:
blocks #17434

PR #358, opened 2017-06-27
Improve number of OS processes detection

mdebruijne:
As discussed in DefinitelyTyped/DefinitelyTyped#17511, this PR improves number of OS processes detection on both Travis CI and development systems.

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -27,7 +27,7 @@ export function currentTimeStamp(): string {
 	return moment().format("YYYY-MM-DDTHH:mm:ss.SSSZZ");
 }
 
-export const numberOfOsProcesses = os.cpus().length;
+export const numberOfOsProcesses = process.env.TRAVIS === "true" ? 8 : os.cpus().length * 4;
 
 /** Progress options needed for `nAtATime`. Other options will be inferred. */
 interface ProgressOptions<T, U> {
--- a/src/lib/packages.ts
+++ b/src/lib/packages.ts
@@ -274,8 +274,8 @@ export type DependencyVersion = number | "*";
 
 export interface TypingsDataRaw extends BaseRaw {
 	readonly dependencies: DependenciesRaw;
-	 // These are always the latest version.
-	 // Will not include anything already in `dependencies`.
+	// These are always the latest version.
+	// Will not include anything already in `dependencies`.
 	readonly testDependencies: string[];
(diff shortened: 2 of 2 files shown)


src/util/util.ts
```diff
@@ -27,7 +27,12 @@ export function currentTimeStamp(): string {
 	return moment().format("YYYY-MM-DDTHH:mm:ss.SSSZZ");
 }
 
-export const numberOfOsProcesses = os.cpus().length;
+export const numberOfOsProcesses = (() => {
```
andy-ms: Use a ternary expression

leonard-thieu:
Should be using a multiple of cores available, right? I did some light testing earlier and was still seeing better performance at 4 processes per core.

| Run | --nProcesses 4 | --nProcesses 8 |
| --- | --- | --- |
| 1 | 15 min 12 sec | 9 min 53 sec |
| 2 | 14 min 27 sec | 9 min 59 sec |
| 3 | 13 min 29 sec | 11 min 19 sec |
| **Avg** | 14 min 23 sec | 10 min 24 sec |

src/util/util.ts
mdebruijne: The reason I have done it this way is that newer versions of TypeScript/node typings want don't like this on a potentially undefined variable. For now it is okay and I have changed it to a ternary expression as requested.

mdebruijne:
I can change the default value. How many processes do you want on Travis CI?

leonard-thieu:
Wouldn't you want it to apply outside of Travis too? I didn't test specifically at 1 process per core but there's still significant performance benefits going from 2 to 4.

mdebruijne:
You have already done a great job benchmarking different values so tell me the best value per core and I will fix accordingly.

leonard-thieu:
I'd go with 4.

mdebruijne:
Done.

andy-ms:
Thanks! I'll publish to the `production` branch tomorrow.

andy-ms:
@mdebruijne Before we do that you should sign the CLA.

andy-ms:
@mdebruijne Still waiting for CLA.

mdebruijne:
@andy-ms I have signed the CLA, but the state in this PR is still cla-required. How to proceed?

andy-ms:
I'll just assume @msftclas is having trouble then. The change should be running on Travis now.

Issue #366, opened 2017-07-30
Add inferred dependencies to peerDependencies? 


sthuck:
Hi guys,

I'll start with the example scenario -
my package.json has these dependencies:
```
    "@types/angular": "~1.5.0",
    "@types/angular-mocks": "^1.5.0"
```
and `@types/angular-mocks` package.json looks like this, as it's dependencies are automatically inferred:
```
 "dependencies": {
        "@types/angular": "*"
    },
```
since `@types/angular": "*"` gets resolved to @types/angular version 1.6.23 (in Yarn 0.27.5), I end up with the types for angular 1.6. Obviously this is just an example and will happen with any type definitions that has breaking changes compared to previous type definitions. 

You can see the discussion with yarn guys https://github.com/yarnpkg/yarn/issues/3951, where they explain why '*' is being resolved to latest version and not simply to "any version". They suggest using `peerDependencies` instead, which does seem (to me at least) like the correct thing to do. 

I do understand this has potential to pop up a 'missing peer dependency' error for lots and lots of people.
But if I understand this bug correctly, as time will pass and new type definitions will break previous versions, more yarn users will experience this issue. So I guess this is for your consideration...

cwmoo740:
Ditto for react.

```
"@types/react": "^15.6.0",
"@types/react-dom": "^15.5.0"
```

where "@types/react-dom" has:

```
"@types/react": "*"
```

This results in the structure:

```
node_modules
-@types
--react (15.6.0)
--react-dom
---node_modules
----@types
-----react (15.0.33)
```

@sthuck and the yarn people seemed to indicate that yarn would pick "latest" for a dependency of *, but that doesn't seem to be the case here.

gilamran:
Installing 
`@types/react@^15.6.1`
`@types/react-dom@^15.5.2` (Depends on `@types/react@*`)

Will install nested `node_modules/@types/react-dom/node_modules/@types/react` (version 16, the latest).

This will cause many errors like this one:
```
node_modules/@types/react-dom/node_modules/@types/react/index.d.ts(3422,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'a' must be of type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>', but here has type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>'.
```

cwmoo740:
[...]

PR #371, opened 2017-08-15
Change dependent types to be included as peerDependencies

glentakahashi:
Add all of the typesDependencies to be added as peerDependencies instead of dependencies to fix transitive dependency issues such as in https://github.com/Microsoft/types-publisher/issues/366

Tested locally and got expected results

### DIFF
--- a/src/lib/package-generator.ts
+++ b/src/lib/package-generator.ts
@@ -127,7 +127,7 @@ function addInferredDependencies(dependencies: Dependencies, peerDependencies: D
 		}
 
 		if (!handlesDependency(dependencies) && !handlesDependency(peerDependencies) && allPackages.hasTypingFor(dependency)) {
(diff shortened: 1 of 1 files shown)


henryptung:
Paging @glentakahashi

RyanCavanaugh:
We can't do this. Editors (VS, VS Code, maybe others) run e.g. `npm install @types/leftpad` and expect that they will get a "working" version of the `leftpad` types package as a result. NPM no longer installs peer dependencies automatically, so this would break editors (as well as user expectation that you can `npm install` some types package without further intervention).

glentakahashi:
Hm okay I took a second reading of peerDependencies and now better understand what theyre trying to accomplish. Will close out for now.

@RyanCavanaugh as a sidenote though, do you have any opinions/ideas on how to solve https://github.com/yarnpkg/yarn/issues/3951 and https://github.com/Microsoft/types-publisher/issues/366 then?

PR #372, opened 2017-08-15
Add CODEOWNERS generate script

andy-ms:


src/code-owners.ts
```diff
@@ -0,0 +1,42 @@
...
+	if (!users.length) {
+		return undefined;
+	}
+
+	const path = pkg.subDirectoryPath.padEnd(maxPathLen);
+	return `/${typesDirectoryName}/${path} ${users.map(u => `@${u}`).join(", ")}`;
+}
+
+function getGithubUsername(url: string): string | undefined {
+	const rgx = /^https\:\/\/github.com\/(.*)$/;
```
RyanCavanaugh: Need to make the s in https optional - some people have usernames in this format

RyanCavanaugh: e.g. https://github.com/kota65535/DefinitelyTyped/blob/5910efc9b693ce73c5889a736119222d4b6771f1/types/paper/index.d.ts

andy-ms: Could just lint for it with Microsoft/dtslint#56 ?

PR #388, opened 2017-10-11
Add redux-saga to whitelist dependencies

LKay:
Module `redux-saga` provides its own types definitions which can be used to create definitions for extending modules.


andy-ms:
@RyanCavanaugh I'll publish the change.

PR #390, opened 2017-10-13
Add log4js to whitelist

Loghorn:


andy-ms:
OK, I presume you will depend on this package somewhere in DefinitelyTyped? Can you link to the PR?

Loghorn:
Actually I don't, I was trying to fix an error in `karma` definitions (that depend on `log4js`)  but I found that the new log4js `index.d.ts` is missing what is needed by karma, so I fixed it in another way. Sorry about that.

PR #395, opened 2017-10-25
Dedupe files entries for index.d.ts

andy-ms:
We were getting an error on Azure: `Error: EPERM: operation not permitted, open 'D:\home\site\wwwroot\output\nodemailer\index.d.ts`. (#394 didn't solve it.)
This was because in `nodemailer`, `index.d.ts` is referenced from a file in a child directory. This caused us to add a file entry for `./index.d.ts`, but there was already one for `index.d.ts`.
Normally this is not a big deal. But since we do our I/O in parallel, during package generation we would try copying `./index.d.ts` and `index.d.ts` at the same time. On Windows this results in `EPERM` for some reason.

### DIFF
--- a/src/lib/module-info.ts
+++ b/src/lib/module-info.ts
@@ -8,10 +8,6 @@ import { hasWindowsSlashes, joinPaths, normalizeSlashes, sort } from "../util/ut
 import { readFileAndThrowOnBOM } from "./definition-parser";
 
 export default async function getModuleInfo(packageName: string, directory: string, allEntryFilenames: string[], log: Logger): Promise<ModuleInfo> {
-	let hasUmdDecl = false;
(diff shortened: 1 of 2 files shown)


src/lib/module-info.ts
```diff
@@ -8,10 +8,6 @@ import { hasWindowsSlashes, joinPaths, normalizeSlashes, sort } from "../util/ut
 import { readFileAndThrowOnBOM } from "./definition-parser";
 
 export default async function getModuleInfo(packageName: string, directory: string, allEntryFilenames: string[], log: Logger): Promise<ModuleInfo> {
-	let hasUmdDecl = false;
```
andy-ms: These are unused variables -- ts2.6 catches these now. (Microsoft/TypeScript#17752)

PR #399, opened 2017-11-06
Added mysql to dependenciesWhitelist.txt 

move-zig:
for use in promise-mysql typings

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -15,6 +15,7 @@ meteor-typings
 mqtt
 moment
 monaco-editor
+mysql
 parse5
 path-to-regexp
 pkcs11js


andy-ms:
`mysql` doesn't seem to come bundled with typings.
There is a `@types/mysql`. If you want to depend on that you don't need a `package.json` at all, just import it.

move-zig:
I believe this is needed for my pull request https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21290 to solve issue https://github.com/DefinitelyTyped/DefinitelyTyped/issues/21100.

andy-ms:
`mysql` doesn't provide its own types. It doesn't make sense to depend on it. You should remove the `package.json` from that pull request.

move-zig:
@andy-ms Thanks!

PR #402, opened 2017-11-14
Add mongodb to dependency whitelist

simonxca:
@types/mongoose require @types/mongodb@2.1.43 or later so we need to add to this whitelist.

This issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/17128
suggests that `@types/mongoose` requires `@types/monogodb@2.1.43` or higher because we require an interface called `Decimal128`.

The Travis build failed telling me to submit a PR here. Please advise. 
https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/301786654?utm_source=github_status&utm_medium=notification

Thank you!

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -15,6 +15,7 @@ meteor-typings
 mqtt
 moment
 monaco-editor
+mongodb
 parse5
 path-to-regexp
 pkcs11js


andy-ms:
You don't need to add something to `dependenciesWhitelist` to rely on an `@types` package.

simonxca:
@andy-ms What should the package.json look like if an `@types` package depends on a specific version of another `@types` package?
https://github.com/simonxca/DefinitelyTyped/blob/cf7854f5b86b89743ab7d0b02a774da0152125c8/types/mongoose/package.json

Should it be:
```js
{
    "private": true,
    "dependencies": {
        "@types/mongodb": ">=2.1.43"
    }
}
```

andy-ms:
@simonxca We really want to avoid that since most people will want to have the latest `@types/mongodb` installed. Can they latest versions of these packages be made compatible?

simonxca:
@andy-ms Oh yeah, the latest versions are compatible. I am actually trying to specify the minimum version of `@types/mongodb`. How would I do that?

andy-ms:
Sorry, we don't really support doing that -- we only generate `"foo": "*"` dependencies. Maybe we should make `types-publisher` always specify a minimum version of whatever the latest is at the time it's published, since that's what we tested against and it remains future-compatible. @RyanCavanaugh @mhegazy thoughts?

simonxca:
Oh I see, I'll remove then.

PR #405, opened 2017-11-21
Add apollo-client to dependenciesWhitelist

pselden:
Needed for this PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21581

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -1,6 +1,7 @@
 activex-helpers
 ajv
 anydb-sql
+apollo-client
 aws-sdk
 axios
 chalk


pselden:
I'm a little skeptical that this is the correct way to go about this, given that the whitelist is so small. apollo-client is written in TypeScript, but mobx-apollo is in JavaScript, and the mobx-apollo maintainer does not use TypeScript and thus would prefer that definitions go in DefinitelyTyped instead.

Is it really the case that any types that depend on external dependencies that publish their own types have to be whitelisted like this? If so, I would have expected this list to be much larger.

PR #408, opened 2017-12-06
Run definition parser in parallel

andy-ms:
This cuts the time by about half on my machine.

### DIFF
--- a/src/make-server-run.ts
+++ b/src/make-server-run.ts
@@ -14,7 +14,7 @@ if (!module.parent) {
 		throw new Error("Must provide PORT");
 	}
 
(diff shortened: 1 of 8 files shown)


src/make-server-run.ts
```diff
@@ -14,7 +14,7 @@ if (!module.parent) {
 		throw new Error("Must provide PORT");
 	}
 
-	const url = remote ? "http://types-publisher.azurewebsites.net" : `http://localhost:${port}`;
+	const url = remote ? "http://typespublisher.azurewebsites.net" : `http://localhost:${port}`;
```
RyanCavanaugh: ?

andy-ms: We moved the server recently and the url changed.

PR #425, opened 2018-01-11
Publish dist tags in types-registry

uniqueiniquity:
In order to support https://github.com/Microsoft/TypeScript/pull/21004 in being able to update acquired types, this change adds information about the most up-to-date version of each package for each version of tsserver. It does this by pulling the `dist-tags` field from npm for each package.


uniqueiniquity:
@mhegazy @RyanCavanaugh @andy-ms

src/lib/versions.ts
```diff
@@ -178,6 +178,7 @@ export interface ProcessedNpmInfo {
 	readonly version: Semver;
 	readonly highestSemverVersion: Semver;
 	readonly contentHash: string;
+	readonly distTags: { [tag: string]: string };
```
andy-ms: Doesn't appear to be used

src/publish-registry.ts
```diff
@@ -152,11 +152,13 @@ function generatePackageJson(version: string, typesPublisherContentHash: string)
...
 
-interface Registry { readonly entries: { readonly [packageName: string]: 1 }; }
-function generateRegistry(typings: ReadonlyArray<TypingsData>): Registry {
-	const entries: { [packageName: string]: 1 } = {};
-	for (const { name } of typings) {
-		entries[name] = 1;
+interface Registry { readonly entries: { readonly [packageName: string]: { readonly [distTags: string]: string } }; }
+async function generateRegistry(typings: ReadonlyArray<TypingsData>): Promise<Registry> {
+	const entries: { [packageName: string]: { [distTags: string]: string } } = {};
+	await nAtATime(25, typings, getTags);
```
andy-ms: Personally I prefer to write just `await nAtATime(25, typings, async typing => { ... })` with no function name or explicit type annotation.

src/lib/versions.ts
uniqueiniquity: Which part is not used? `distTags` is needed for line 191.

EDIT: Never mind I see what you mean.

PR #429, opened 2018-01-24
Reduce size of types-registry

uniqueiniquity:
#425 increased the number of lines in types-registry by a factor of 10 since a version was listed for every version of typescript (plus latest) for every package.

This change removes all whitespace from the registry and only lists tags that differ from the latest version.

### DIFF
--- a/src/publish-registry.ts
+++ b/src/publish-registry.ts
@@ -32,7 +32,7 @@ export default async function main(options: Options, dry: boolean): Promise<void
 
 	// Don't include not-needed packages in the registry.
 	const typings = await AllPackages.readTypings();
-	const registry = JSON.stringify(await generateRegistry(typings), undefined, 4);
+	const registry = JSON.stringify(await generateRegistry(typings), undefined);
 	const newContentHash = computeHash(registry);
 
 	assert.equal(oldVersion.major, 0);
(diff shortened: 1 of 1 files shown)


uniqueiniquity:
@mhegazy @andy-ms could you take a look?

src/publish-registry.ts
```diff
@@ -32,7 +32,7 @@ export default async function main(options: Options, dry: boolean): Promise<void
 
 	// Don't include not-needed packages in the registry.
 	const typings = await AllPackages.readTypings();
-	const registry = JSON.stringify(await generateRegistry(typings), undefined, 4);
+	const registry = JSON.stringify(await generateRegistry(typings), undefined);
```
mhegazy: you do not need the second `undefined` parameter either.

andy-ms:
Could we just leave it as `1` if the latest tag is the only one? `npm install` installs the latest by default without needing a version number. Also, we don't update `types-registry` at every package update, just on new packages, so it would get out of date.

src/publish-registry.ts
```diff
@@ -157,7 +157,21 @@ async function generateRegistry(typings: ReadonlyArray<TypingsData>): Promise<Re
 	const entries: { [packageName: string]: { [distTags: string]: string } } = {};
 	await nAtATime(25, typings, async typing => {
 		const info = await fetchNpmInfo(typing.fullEscapedNpmName);
-		entries[typing.name] = info["dist-tags"];
+		const tags = info["dist-tags"];
+		if (tags) {
```
andy-ms: This should always be defined.

uniqueiniquity:
I use the version in the registry to decide in the installer whether or not I need to update the installed copy. So I don't think keeping the `1` is possible.

PR #432, opened 2018-02-01
chore(DependenciesWhitelist): add D3 to Whitelist to prevent breaking changes

tomwanzek:
@andy-ms this is the initial PR to add `d3` for `package.json`-stubs in line with our discussion here: DefinitelyTyped/DefinitelyTyped#23257.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -8,6 +8,7 @@ chalk
 cordova-plugin-camera
 cordova-plugin-file
 cordova-plugin-file-transfer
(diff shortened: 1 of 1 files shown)


andy-ms:
Don't think this will be necessary if you use path mapping, see my comment on that issue.

tomwanzek:
I will respond back in the main issue thread on DT.

tomwanzek:
Closed as per discussion in DefinitelyTyped/DefinitelyTyped#23257

PR #433, opened 2018-02-03
Add csstype to whitelist

pelotom:


andy-ms:
Just to confirm, you're planning on importing from `csstype` in the linked PR? It looks like currently you're not using it.

pelotom:
Yes, planning to.

PR #445, opened 2018-03-22
Quick fix for missing "deprecated" property from npm info

andy-ms:
Npm isn't reliably returning this property -- I'll look into getting a proper repro and filing a bug there, but this will get the publisher working without extra publishes for now.


andy-ms:
The problem appears to be fixed after I un-deprecated and re-deprecated everything.

PR #449, opened 2018-04-08
Redue noise in output by not printing OK for happy paths

mohsen1:
A better solution is to use a debugger like [debug](https://www.npmjs.com/package/debug) to be able to turn on those loggings.


RyanCavanaugh:
Thanks @mohsen1 - more like this definitely (ha!) welcomed!

andy-ms:
This may be causing our builds to time out because of no output received for 10 minutes. https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

PR #450, opened 2018-04-14
Add type-zoo to whitelist

pelotom:
There are a few utility types that end up getting implemented over and over again from scratch in DefinitelyTyped packages. This package obviates the need for that:

https://github.com/pelotom/type-zoo

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -47,6 +47,7 @@ smooth-scrollbar
 source-map
 styled-components
 typescript
+type-zoo
 tweetnacl
 vue
 vuex


andy-ms:
The library seems to duplicate some times already in ts2.8, but with different names. For example, `Diff` becomes `Exclude`. Might be better to just wait for ts2.8 to be available on DT?

pelotom:
Yes, some but not all of its types will be superseded when 2.8 arrives. But
1. It still has other useful types such as `Omit` which are not covered by the stdlib
2. Many libraries will want to maintain backwards compatibility with earlier versions of TS as long as possible. As long as they don’t directly need the full power of conditional types themselves, they won’t want to make 2.8 their baseline requirement

andy-ms:
The reason we didn't have some of those types in the standard library before is that the workarounds without contextual types didn't work well in general. For example, `lowdb` is broken right now (Microsoft/TypeScript#22603), and we'll probably have to convert it to using real contextual types to fix. @sandersn @RyanCavanaugh thoughts?

Issue #451, opened 2018-04-18
Improve log readability of test output


RyanCavanaugh:
Today when a Travis build fails, e.g. https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/362263112?utm_source=github_status&utm_medium=notification , the actual error introduced by the change can be extremely low in the file (in the linked one, it's around line 4,800).

Ideally we would just not have quite so much log spam (all the "duplicate" notifications, for example, are useless) and the error could be up top. If there's actual diagnostic value in some of them, though, there appears to be log folding support: https://github.com/travis-ci/travis-ci/issues/2285

I think the general goal should be that we can see the relevant error text without scrolling down more than a screenful on the Travis webpage - DT authors will probably at least scroll down this far and realize what they've broken

mhegazy:
https://github.com/Microsoft/types-publisher/pull/449 removes the `<package> OK` line.

PR #452, opened 2018-04-25
Print 'ok' when a test passes

andy-ms:
Reverts #449
May fix the timeout in DefinitelyTyped/DefinitelyTyped#25269

### DIFF
--- a/src/tester/test-runner.ts
+++ b/src/tester/test-runner.ts
@@ -94,7 +94,9 @@ async function runTests(
 		cwd: options.typesPath,
 		handleOutput(output): void {
 			const { path, status } = output as { path: string, status: string };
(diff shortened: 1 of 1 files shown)


mohsen1:
Can you print a dot with no new lines to avoid long build logs?

andy-ms:
I think the real problem is the many duplicate packages -- see #451. We should invest some effort into genuinely cleaning up DT rather than trying to hide the problems...

mhegazy:
@andy-ms we can use travis folding APIs and generated fordable sections around that. most ppl do not have to see them this way.

we use them in TS build, see https://github.com/Microsoft/TypeScript/blob/master/Jakefile.js#L636-L643

PR #453, opened 2018-04-25
Use travis-fold

andy-ms:
Fixes https://github.com/Microsoft/types-publisher/pull/452#issuecomment-384450246


src/types/travis-fold.d.ts
```diff
@@ -0,0 +1,8 @@
+declare module "travis-fold" {
```
andy-ms: Also made DefinitelyTyped/DefinitelyTyped/pull/25305

PR #466, opened 2018-06-04
Add ethereum-types to the whitelist

LogvinovLeon:
Ethereum ecosystem has a lot of packages that share some types that are not actually related to any package, but rather to ethereum in general. I'm planning to add a bunch of typings to DT in the future. I already have definitions ready. [ethereum-types](https://www.npmjs.com/package/ethereum-types) is the package that contains those shared types and will be used for DT definitions.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -13,6 +13,7 @@ csstype
 decimal.js
 egg
 electron
+ethereum-types
 eventemitter2
 eventemitter3
 fast-glob


andy-ms:
If this package will only exist for use by DT packages, why not just make `@types/ethereum`?

LogvinovLeon:
Because ethereum npm name is taken by some unpopular ethereum node implementation and those types do not relate to that https://www.npmjs.com/package/ethereum.

LogvinovLeon:
People might think that those definitions are the definitions for that package when in reality they're generic definitions. Think about it that way. Where should we put the DOM definitions? In @types/dom? Will make sense if someone will just block dom name on npm, but here `ethereum` name is already takes.

andy-ms:
There are DT packages that are related to a package without being for that package itself, such as `express-serve-static-core`. We might consider an `ethereum-core` package that's used by `ethereum` as well as other related packages. Another alternative would be to just put all the definitions in `@types/ethereum` and have other typings just take a direct dependency on that.

LogvinovLeon:
So I should take an npm name `ethereum-core`. Put a placeholder there so that noone takes it. Submit PR to DT with `@types/ethereum-core` typings. Close this PR. Use @types/ethereum-core as a dependency?

andy-ms:
Could you explain how the types are related to ethereum but don't belong in `@types/ethereum`?

LogvinovLeon:
I'll bring the analogy with WEB.

**Ethereum** = **Internet**
**Chrome** (as some particular browser) = **npm ethereum package**
**DOM, Window, Document** (as some web related types = **ethereum types**.

Let's say someone deploys `Chrome` source code on npm and calls the package `internet`.
Where should we put the `DOM` types then? We could put them in `@types/internet`, but people will think that those are types for internet - the package, not internet - the protocol. The problem here is that someone squatted the name on npm and put a specific package under the name of a protocol.

The similar thing happened to the dom itself. npm name dom points to a particular library to manipulate dom https://www.npmjs.com/package/dom. Why don't we have all the DOM types in @types/dom then?

andy-ms:
I think a `@types/ethereum-protocol` package would be reasonably safe to add.

LogvinovLeon:
@andy-ms I'll do that. Will I be able to reference it as a dependency then without adding to this package? If so - I'll close this PR. Thanks for guidance either way. Appreciate that!

andy-ms:
Right, you don't even need `package.json` for inter-DT dependencies.

LogvinovLeon:
BTW, those types depend on `bignumber.js` types. I can't figure out how to properly import them. Do I need to submit a PR to this repo adding `bignumber.js` to whitelist?

andy-ms:
Yeah, for that one you need to add it to the list.

PR #470, opened 2018-06-26
Reduce number of processes on travis to 2

Flarna:
Reduce the number of processes in travis environment to 2 as current setup is CPU bound and travis provides just two CPUs (see https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System).

Using more processes results only in higher memory consumption with the risk to hit the 4GB limit of travis containers.

fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/26063


andy-ms:
Thanks!

PR #471, opened 2018-07-04
Fix "stdout maxBuffer exceeded" error

tkrotoff:
Increases `exec()` buffer size from 200 KiB to 1 MiB

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26545#issuecomment-402274021


andy-ms:
Thanks!

PR #480, opened 2018-08-24
Whitelist Reselect and packages from @planttheidea

nickmccurdy:
All these packages are distributed with index.d.ts files. This should bring the whitelist up to date for all packages from @reduxjs and @planttheidea.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -19,7 +19,10 @@ egg
 electron
 eventemitter2
 eventemitter3
+fast-copy
+fast-equals
 fast-glob
+fast-stringify
 flatpickr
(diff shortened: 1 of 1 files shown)


andy-ms:
@nickmccurdy Do you need to use these packages as dependencies of DefinitelyTyped packages? If not, there's no reason to add them to the whitelist. If so, please link to the DefinitelyTyped PRs affected.

nickmccurdy:
That was my original intent, but since the package author already has some packages with declaration files published, I'm thinking it's better to send a pull request to the original repository instead of DefinitelyTyped. I'll reopen this if I decide to use DefinitelyTyped instead.

Issue #485, opened 2018-09-26
Are types being published right now?


johnnyreilly:
Are types being published right now? My PR was merged yesterday but there's no sign on npm:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29186

https://www.npmjs.com/package/@types/recharts

https://www.npmjs.com/package/@types/recharts-scale

More generally, it would be great if we could get some visibility on the types publishing process; at present there's no information available between merging and publishing which can lead the puzzlement! This had been mentioned previously but I'm not aware of any remedies.

Interestingly in the linked issue a key had expired. That was 12 months ago - I can't help but wonder if the same might have happened?

https://github.com/Microsoft/types-publisher/issues/384

RyanCavanaugh:
#40 tracks server status. @andy-ms can you investigate?

johnnyreilly:
Sorry to chase - just wondering if there was any news on this?

johnnyreilly:
Looks like we're back! Thanks!

PR #486, opened 2018-09-26
On azure, clone DT from a zip

andy-ms:
We've had many problems in the past due to the DefinitelyTyped repository on azure getting corrupted. This would make it so that there's no longer any state that needs to be kept between types-publisher runs (though keeping the `cache` directory around improves performance) -- we always get a fresh ZIP file instead. On my laptop this takes under 30 seconds.


johnnyreilly:
If merged will this resolve https://github.com/Microsoft/types-publisher/issues/485?

https://www.npmjs.com/package/@types/recharts is still not published; I imagine others are affected too

PR #502, opened 2018-10-17
Fix scoped at types dependencies

### LINKED ISSUE #498: Generated package.json missing package dependencies for scoped packages
I did some investigating myself because I really need https://github.com/DefinitelyTyped/DefinitelyTyped/issues/27729 to get resolved.

The bug occurs when types for a scoped package contain dependencies on other scoped packages. When collecting the dependencies for a generated `package.json` here...
https://github.com/Microsoft/types-publisher/blob/745ed6d124dbb5c8e755beb2aa88130b988c9c62/src/lib/package-generator.ts#L74

...it walks over the dependencies specified by the typing...
https://github.com/Microsoft/types-publisher/blob/745ed6d124dbb5c8e755beb2aa88130b988c9c62/src/lib/package-generator.ts#L114

...but when calling `allPackages.hasTypingFor`, the argument contains the scoped (unmangled) package name but it expects a mangled package name when called here...

https://github.com/Microsoft/types-publisher/blob/745ed6d124dbb5c8e755beb2aa88130b988c9c62/src/lib/package-generator.ts#L122

...therefore not adding the dependency because for example calling `allPackages.hasTypingFor("@babel/template")` will return `false` but `allPackages.hasTypingFor("babel__template")` return will return `true`.

~~I did not create a PR because I do not have a good enough understanding of the codebase but if it's a simple change I can still make one if requested.~~ I gave it a try anyway. See referenced PR.

mgroenhoff:
Fixes #498

Also:

* Made sure dependencies are always ordered alphabetically in generated package.json
* Fixed an issue where dependencies listed in readme.md would not contain the correct package name.
* Also fixed an issue where an `NpmInfoClient` would return something without a `dist-tags` or a `versions` property (Happens with unpublished packages. See http://registry.npmjs.org/emscripten).
* Fixes an issue where parsing definitions for a single package using `--single=package_name` always generated an empty `definitions.json` file.

For reference: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/27729


mgroenhoff:
@andy-ms Can you please review this? This is the first time I touched the code of `types-publisher` so its possible that I might have missed things.

Also when this is merged and published/used, how do I make sure that some packages are published without making changes to all of them?

src/lib/npm-client.ts
```diff
@@ -92,6 +92,10 @@ export class UncachedNpmInfoClient {
 			if (info.error === "Not found") { return undefined; }
 			throw new Error(`Error getting version at ${escapedPackageName}: ${info.error}`);
 		}
+		if (!info["dist-tags"] && !info.versions) {
```
andy-ms: When does this happen? I thought we get "error" if the package doesn't exist. If the package does exist it should have "versions".

src/lib/packages.ts
```diff
@@ -92,7 +92,7 @@ export class AllPackages {
 	}
 
 	tryGetTypingsData({ name, majorVersion }: PackageId): TypingsData | undefined {
```
andy-ms: We should clarify what a `PackageId` is. The interface is possibly misnamed since it only refers to `@types` packages and not any npm package. As of this PR, it looks like `name` in PackageId should always be `@foo/bar` not `foo__bar`? Might want to convert it to a class so we could assert that in the constructor.

andy-ms:
`npm run calculate-versions` also calculates which packages are to be published. So if this change caused unnecessary extra publishes you could tell from that.

src/lib/npm-client.ts
mgroenhoff: Yeah I thought so too but when I was trying to generate all packages it was trying to get the package info for `emscripten` which had *no* `versions`. See http://registry.npmjs.org/emscripten

mgroenhoff:
I am getting a new error when I run `npm run generate -- --all`
```
Error: No version info for axe-core@3
    at Versions.info (C:\Source\types-publisher\src\lib\versions.ts:140:10)
    at Versions.getVersion (C:\Source\types-publisher\src\lib\versions.ts:129:48)
    at generateNotNeededPackage (C:\Source\types-publisher\src\lib\package-generator.ts:40:63)
    at Object.generateAnyPackage [as default] (C:\Source\types-publisher\src\lib\package-generator.ts:16:29)
    at util_1.nAtATime (C:\Source\types-publisher\src\generate-packages.ts:41:40)
    at Promise.all.initArray (C:\Source\types-publisher\src\util\util.ts:60:25)
```

I think wat it tries to do is this:
* `node bin/calculate-versions.js` generates `versions.json` and contains this piece:

  ```json
  "axe-core": {
      "2": {
          "patch": 7
      }
  }
  ```
* `node bin/generate-packages.js` tries to generate a "not-needed" package for the deprecated package `axe-core` because it is in DT's `notNeededPackages.json`:

  ```json
  {
      "libraryName": "axe-core",
      "typingsPackageName": "axe-core",
      "sourceRepoURL": "https://github.com/dequelabs/axe-core",
      "asOfVersion": "3.0.3"
  }
  ```
* When it gets to here and calls `versions.getVersion(pkg)` here:
https://github.com/Microsoft/types-publisher/blob/42307e76c34c7e0de3e4d8fd36d4f93c9ee37a75/src/lib/package-generator.ts#L40

  `versions.getVersion(pkg)` calls `versions.info(pkg.id)` and comes down to this: `pkg.id` contains `{ name: "axe-core", majorVersion: 3 }` and `versions.data["axe-core"]` contains `{ "2": { patch: 7 } }` which causes `versions.info()` to throw with the error above because it was never put into `versions.json`. The result of `versions.getVersion(pkg)` would have been places in the final `version` property of the `package.json`.

I don't think my changes cause this but is this a bug?

src/lib/packages.ts
mgroenhoff: What should these type reference directives contain? `@foo/bar` or `foo__bar`?
https://github.com/Microsoft/types-publisher/blob/42307e76c34c7e0de3e4d8fd36d4f93c9ee37a75/src/lib/module-info.ts#L31-L33

Docs at https://github.com/DefinitelyTyped/DefinitelyTyped#im-writing-a-definition-that-depends-on-another-definition-should-i-use-reference-types--or-an-import does not mention that

mgroenhoff: I found some inconsistencies while trying figure this out:
* some packages contain the imports like `import ... from "foo__bar"` instead of `import ... from "@foo/bar"`
  * `@types/material-components-web`
  * `@types/material__*`
  * `@types/ember`
* some packages contain types references like `<reference types="foo_bar" />` instead of `<reference types="@foo/bar" />`
  * `ember`
  * `feathersjs__primus` and `feathersjs__socketio`

All this causes `getModuleInfo()` return an inconsistent set of dependencies to begin with.
https://github.com/Microsoft/types-publisher/blob/42307e76c34c7e0de3e4d8fd36d4f93c9ee37a75/src/lib/module-info.ts#L10

mgroenhoff: @andy-ms Should I just leave this as it is and maybe create a new issue for this so the rest can be merged as it fixes the original problem?

mgroenhoff:
@andy-ms Can you take a look at this again please?

andy-ms:
Pinging @sandersn to take a look when he's back.

mistic:
@mgroenhoff  @andy-ms @sandersn this will impact even more people very soon, can we make an effort to push this further?

mgroenhoff:
@andy-ms @sandersn I am trying to get the conflicts resolved and I wanted to generate all packages to see if it still works but could it be that the documented '--forceUpdate' flag to 'node bin/calculate-versions.js' no longer works? I also don't see any references in the code.

sandersn:
@mgroenhoff It does look like there's some maintenance needed. For the last week I've been added tests, logging and instrumentation so I will know when I've broken something. Feel free to send PRs to eliminate unused features, or I'll get around to it eventually.

Anyway, thanks for your set of fixes here!

mortyccp:
@sandersn I encounter the same problem when using https://www.npmjs.com/package/@types/storybook__addon-storyshots. What can I do to resolve this problem? Anyway I can trigger a publish for this package?

sandersn:
Probably the easiest way is to make a trivial edit to the package. I can do that today when I get off the bus.

sandersn:
The PR is here: 
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31547

I’ll merge it after CI passes.

sandersn:
@mortyccp Publishing is finished. I see an additional dependency, @types/storybook__react. Is that correct?

PR #507, opened 2018-10-29
fix path mapping checks for scoped and versioned packages

### LINKED ISSUE #506: Scoped & versioned packages can't pass path mappings
In https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29979, I was trying to create a new major version of a scoped package. In order to do that, you're supposed to add a path mapping similar to the example in the DefinitelyTyped readme:

```
        "paths": {
            "history": [ "history/v2" ]
        },
```

For the package I was trying to update, I believe that'd be:

```
        "paths": {
            "@expo/vector-icons": [ "expo__vector-icons/v6" ]
        },
```

As far as I can tell, [the code here](https://github.com/Microsoft/types-publisher/blob/10294c0e763cc63a5503f5782f31563ed7f80e71/src/lib/definition-parser.ts#L285-L292) enforces that the mapping instead be exactly:

```
        "paths": {
            "@expo/vector-icons": [ "expo__vector-icons" ]
        },
```

Would a PR be accepted to change this behavior?

JKillian:
Fixes #506 

Fix how we parse path mappings for scoped and version packages.

Disclaimer: I wasn't sure how to fully test out my changes, please feel free to give things a run through. I did test it on the changes from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29979 though and it seemed to resolve the issues.

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -3,7 +3,7 @@ import * as ts from "typescript";
 
 import { FS } from "../get-definitely-typed";
 import {
-	computeHash, filter, flatMap, hasWindowsSlashes, join, mapAsyncOrdered, mapDefined, split, unique, unmangleScopedPackage, withoutStart,
+	computeHash, filter, flatMap, hasWindowsSlashes, join, mapAsyncOrdered, mapDefined, split, stripVersion, unique, unmangleScopedPackage, withoutStart,
 } from "../util/util";
 
 import getModuleInfo, { getTestDependencies } from "./module-info";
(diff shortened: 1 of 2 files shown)


src/lib/definition-parser.ts
```diff
@@ -282,8 +282,8 @@ async function calculateDependencies(
 		}
 		const pathMapping = pathMappingList[0];
 
-		// Path mapping may be for "@foo/bar" -> "foo__bar".
-		const scopedPackageName = unmangleScopedPackage(pathMapping);
+		// Path mapping may be for "@foo/bar" -> "foo__bar" or "@foo/bar" -> "foo__bar/v2"
+		const scopedPackageName = unmangleScopedPackage(stripVersion(pathMapping));
 		if (scopedPackageName !== undefined) {
 			if (dependencyName !== scopedPackageName) {
 				throw new Error(`Expected directory ${pathMapping} to be the path mapping for ${dependencyName}`);
```
sandersn: `calculateDependencies` is odd in that only versioned packages need an entry in the dependencies it calculates. Previously, since we didn't check scoped packages for versions, the `continue` below was fine.

However, if you want to correctly handle versioned, scoped packages, then you'll have to move this section below the version-checking code that starts with `const majorVersion = ...` on line 294. Then that section will need to handle scoped packages correctly. I guess that the handling will be something like `{ name: scopedPackageName, majorVersion }`, where `scopedPackageName = unmangleScopedPackage(stripVersion(pathMapping))` (like you have above) and `majorVersion = parseDependencyVersionFromPath(...`. But you'll have to make sure that parseDependencyVersionFromPath handles scoped packages as well, perhaps by calling it with different arguments for the scoped version case.

PR #509, opened 2018-11-03
Fixes to publish-registry

andy-ms:
* Reuse `allPackages` instead of loading it again
* Create `publishClient` lazily (allow to run most of this without credentials)
* Use `allLatestTypings()` instead of `allTypings()`, since registry only stores info for latest version
* Publish if it *has* been a week, not hasn't

Also, there were validation failures before because we started sorting packages; did a new publish and validation now passes.


src/publish-registry.ts
```diff
@@ -20,20 +20,20 @@ Generated by [types-publisher](https://github.com/Microsoft/types-publisher).`;
 
 if (!module.parent) {
 	const dry = !!yargs.argv.dry;
-	done(getDefinitelyTyped(Options.defaults).then(dt => main(dt, dry, new UncachedNpmInfoClient())));
+	done(async () => {
+		const dt = await getDefinitelyTyped(Options.defaults);
+		await main(dt, await AllPackages.read(dt), dry, new UncachedNpmInfoClient());
```
sandersn: since I'm not familiar with async syntax, I want to confirm: the only substantive change here is to add a call to `AllPackages.read(dt)`, right? I think everything else is a switch from `.then(dt =>` ... to `dt = await ...`, right?

PR #511, opened 2018-11-13
Added @sentry/browser package

dolezel:


dolezel:
Because of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30490

andy-ms:
It looks like that PR only uses `@sentry/browser` -- I'd like to keep the list as small as possible so could you limit the PR to just that?

PR #519, opened 2018-11-16
Assert registry directory is subset, not deep equal

### LINKED ISSUE #518: publish-registry validation fails when new packages are out but we won't publish them for a week
To avoid excessive publishing, publish-registry only publishes if it's been a week since the last publish. Unfortunately it then asserts that the registry on npm is *exactly* what we have locally, which it won't be if new packages have been added. Instead we should always validate that the one on npm is a subset of the local one.

sandersn:
Fixes #518

### DIFF
--- a/src/util/io.ts
+++ b/src/util/io.ts
@@ -138,22 +138,23 @@ export async function isDirectory(path: string): Promise<boolean> {
 	return (await stat(path)).isDirectory();
 }
 
-export async function assertDirectoriesEqual(expected: string, actual: string, options: { ignore(fileName: string): boolean }): Promise<void> {
+/** All entries in `actual` must appear in `expected`. If an `actual` entry is a directory, the `expected` must be too.  */
+export async function assertDirectoryIsSubset(expected: string, actual: string, options: { ignore(fileName: string): boolean }): Promise<void> {
 	const expectedLs = await readdir(expected);
 	const actualLs = await readdir(actual);
-	assert.deepStrictEqual(expectedLs, actualLs);
-	for (const name of expectedLs) {
+	for (const name of actualLs) {
 		if (options.ignore(name)) {
 			continue;
 		}
 
 		const expectedFile = joinPaths(expected, name);
 		const actualFile = joinPaths(actual, name);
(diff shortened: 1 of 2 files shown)


src/util/io.ts
```diff
@@ -138,22 +138,23 @@ export async function isDirectory(path: string): Promise<boolean> {
...
 		const actualFile = joinPaths(actual, name);
+		assert(expectedLs.indexOf(name) > -1, `Unexpected entry ${name} found in ${actualFile} but not ${expectedFile}.`);
 		const expectedStat = await stat(expectedFile);
 		const actualStat = await stat(actualFile);
 		assert.strictEqual(expectedStat.isDirectory(), actualStat.isDirectory());
 		if (expectedStat.isDirectory()) {
-			await assertDirectoriesEqual(expectedFile, actualFile, options);
+			await assertDirectoryIsSubset(expectedFile, actualFile, options);
 		} else {
 			assert.strictEqual(await readFile(actualFile), await readFile(expectedFile));
```
andy-ms: This is still asserting that the file contents are exactly equal. The set of files that exist in `types-registry` doesn't change, but the content of `index.json` does.

sandersn: So is the right fix just to add `"index.json"` to `validate`'s ignore function?

sandersn: Or should I change the function to read specific (JSON) files and pass them on to another function, which then makes a subset assertion about JSON instead of file structure?

andy-ms: `assertDirectoriesEqual` is a red herring because `index.json` is the only really important file in `types-registry`. Though we do want to assert that no extra files somehow made there way there and the `package.json` isn't corrupted somehow.

We should assert that the published `index.json` doesn't contain anything it shouldn't, so its contents should be a subset of what a fully up-to-date `index.json` would contain.

It might also be nice to assert that the published `index.json` does contain everything it should, which means every package more than a week old.

andy-ms: Since `types-registry` only contains `package.json`, `index.json`, and `README.md`, having a recursive `assertDirectoriesEqual` utility may be overkill.

PR #523, opened 2018-11-21
Convert tabs to spaces

sandersn:
UNDER NEW MANAGEMENT


sandersn:
I confirmed that when github's view is switched to Hide Whitespace, all files have only whitespace changes.

PR #543, opened 2018-12-07
Add gitHead to package.json

### LINKED ISSUE #458: [request] Append `gitHead` to package.json
Currently published @​types packages don't include git revision of DefinitelyTyped repo, so we cannot get to know the exact diff between previous version and new one without actually comparing downloaded packages.

`npm publish` appends `gitHead` field to package.json: https://github.com/npm/read-package-json/issues/7

I think the field is very useful for @​types packages.

felixfbecker:
This is a field normally added automatically by npm publish.

Closes #458

### DIFF
--- a/src/get-definitely-typed.ts
+++ b/src/get-definitely-typed.ts
@@ -8,8 +8,9 @@ import * as zlib from "zlib";
 
 import { Options } from "./lib/common";
 import { dataDirPath, definitelyTypedZipUrl } from "./lib/settings";
-import { readFile, readJson, stringOfStream } from "./util/io";
-import { assertDefined, assertSorted, Awaitable, exec, joinPaths, withoutStart, logUncaughtErrors } from "./util/util";
+import { Fetcher, readFile, readJson, stringOfStream } from "./util/io";
+import { assertDefined, assertSorted, Awaitable, exec, joinPaths, logUncaughtErrors, withoutStart } from "./util/util";
+import { queryGithub } from "./util/github";
 
 /**
  * Readonly filesystem.
@@ -43,10 +44,18 @@ if (!module.parent) {
     });
 }
 
-export async function getDefinitelyTyped(options: Options): Promise<FS> {
+/**
+ * Return the commit SHA1 of the master tip of DefinitelyTyped
+ */
+export async function resolveDefinitelyTypedMaster(githubAccessToken: string, fetcher: Fetcher): Promise<string> {
+    const masterRef = await queryGithub("repos/DefinitelyTyped/DefinitelyTyped/git/refs/heads/master", githubAccessToken, fetcher) as { object: { sha: string } };
+    return masterRef.object.sha
+}
+
+export async function getDefinitelyTyped(options: Options, revision: string = 'master'): Promise<FS> {
     if (options.definitelyTypedPath === undefined) {
         await ensureDir(dataDirPath);
-        return downloadAndExtractFile(definitelyTypedZipUrl);
+        return downloadAndExtractFile(definitelyTypedZipUrl + "/" + revision);
     } else {
         const { error, stderr, stdout } = await exec("git diff --name-only", options.definitelyTypedPath);
         if (error) { throw error; }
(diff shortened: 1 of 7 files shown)


felixfbecker:
@sandersn before I solve the merge conflict, could you maybe take a look at this PR?

src/get-definitely-typed.ts
```diff
@@ -43,10 +44,18 @@ if (!module.parent) {
     });
 }
 
-export async function getDefinitelyTyped(options: Options): Promise<FS> {
+/**
+ * Return the commit SHA1 of the master tip of DefinitelyTyped
+ */
+export async function resolveDefinitelyTypedMaster(githubAccessToken: string, fetcher: Fetcher): Promise<string> {
+    const masterRef = await queryGithub("repos/DefinitelyTyped/DefinitelyTyped/git/refs/heads/master", githubAccessToken, fetcher) as { object: { sha: string } };
```
sandersn: Is master the correct SHA to use here? In particular, it's possible that it may have moved between the time that this package was merged. Another package may have been merged in between.

I think it would be better if you use the latest sha just for the package's directory.

felixfbecker: It doesn't really matter - this SHA will be used to download the zip, so it is guaranteed that `gitHead` points to the exact revision that is being published. That is all that should matter.

It would be nice, but I wouldn't know where to get that commit from. Can you point to where?

src/lib/package-generator.ts
```diff
@@ -52,11 +52,12 @@ async function outputFilePath(pkg: AnyPackage, filename: string): Promise<string
 
 interface Dependencies { [name: string]: string; }
 
-async function createPackageJSON(typing: TypingsData, version: string, packages: AllPackages): Promise<string> {
+async function createPackageJSON(typing: TypingsData, version: string, gitHead: string, packages: AllPackages): Promise<string> {
     // Use the ordering of fields from https://docs.npmjs.com/files/package.json
```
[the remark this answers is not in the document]
felixfbecker: You can find it in the registry API response: http://registry.npmjs.org/commander/latest
And npm should also write it out in node_modules when installing.
`npm info commander gitHead` also returns it (or `npm info commander --json`).

But the order of keys shouldn't really matter in an object.

It gets added by npm here: https://github.com/npm/read-package-json/blob/1b85cf87f5875774e7fbc617f606923d8cbda6a4/read-json.js#L345

src/get-definitely-typed.ts
sandersn: Here’s the code I used, although that was using octokit. The rest call shouldn’t be too different though. 

https://github.com/sandersn/types-publisher-watchdog/commit/f6471155a26702692e8abcd82c001c408e78354c#diff-168726dbe96b3ce427e7fedce31bb0bcL76

Look at the part I deleted starting with `const oops`

sandersn:
I also thought it would be easier for me to understand the change with a test — I added jest a little while ago. If you don’t have time for that I can try to add it on Monday.

src/get-definitely-typed.ts
felixfbecker: Ah, so you would get the latest commit from the git log of that folder. I thought you were suggesting to use the commit ID from the triggering event, e.g. the commit hash of the merge commit of a PR that was merged which triggered the new publish

sandersn:
Here's the test I added. However, I can't get github API requests to work at all (eg from `npm run get-definitely-typed`), so I haven't been able to verify that it passes.

```ts
+    async definitelyTypedMasterResolvesToCommit() {
+        expect(process.env["GH_API_TOKEN"]).toBeDefined();
+        const s = await resolveDefinitelyTypedMaster(process.env["GH_API_TOKEN"]!, new Fetcher());
+        expect(typeof s).toBe("string");
+        expect(s.length).toBe(40);
+        expect(/[^a-z0-9]/.test(s)).toBe(true);
+    }
```

I'll push a commit myself if I can get github API requests working but I don't have any more time today to do so.

PR #568, opened 2019-02-06
Make types-publisher consumable via require

sandersn:
And move npmTags to its own repo, which then consumes types-publisher via require. The exported interface is exactly and only the symbols needed for npmTags right now. I literally replaced 'import' with 'export'.

As part of this:

1. I added index.ts with an associated "main" entry in package.json.
2. I added declaration: true in tsconfig.json.
3. I added a "types" entry for index.d.ts in package.json.
4. I removed npmTags.ts.
5. I moved the worker functions to package-publisher.ts where they are actually used.


sandersn:
The new repo for npmTags is https://github.com/sandersn/dt-retag, at least for now.

PR #571, opened 2019-02-08
Add "handlebars" to the dependenciesWhitelist

nknapp:
Handlebars ships its own types since version 4.1.0

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -44,6 +44,7 @@ google-auth-library
 google-gax
 graphql-tools
 grpc
+handlebars
 immutable
 indefinite-observable
 inversify


nknapp:
Could anybody merge this? The removal of typings from DefinitelyTyped depends on it.

jamescdavis:
Bump. @sandersn?

chriskrycho:
Bump again – this is blocking DefinitelyTyped/DefinitelyTyped#33200, which blocks *every single* Ember types package.

chriskrycho:
Thank you, @sandersn! 😊

sandersn:
Should be published now as well. I restarted the Travis build to see whether it passes now.

sandersn:
Now DefinitelyTyped/DefinitelyTyped#33200 is hitting the typescript@next break for ember, which is tracked at https://github.com/Microsoft/TypeScript/issues/29898. @chriskrycho if you are an ember expert, you may be able to point out the best fix for what is going wrong here.

chriskrycho:
Gah. @mike-north had dug the furthest on that particular one. If he doesn't have a chance, I'll try to swing back around on this later today or tomorrow.

nknapp:
I'm not sure if the ember typings could not be made independent of Handlebars actually. All they need is "SafeString", which is pretty simple.

```
 export class SafeString {
      constructor(str: string);
      toString(): string;
      toHTML(): string;
  }
```

Seems like a lot of effort to depend on the whole Handlebars package just for that.

mike-north:
> I'm not an ember expert, but as far as I know, ember isn't using Handlebars anymore.

chriskrycho:
@nknapp Ember does still use (a descendant of) Handlebars which uses Handlebars' types and AST. It still might be worth our extracting `SafeString` from that to a shared location, though. 🤔

nknapp:
I think that ember's decendant of Handlebars (you mean Glimmer?) ist not compatible to Handlebars 4.x though, but rather to 3.x

Now, the typing packages should somehow resemble to real package-structure. Which package do they get the SafeString class from? 

Just that I understand correctly: When my PR is merged, and somebody installs `@types/ember`, does that pull `handlebars@4.1.0` into its dependencies as well?

nknapp:
After reading the comments on https://github.com/wycats/handlebars.js/issues/1499 and https://github.com/ef4/ember-auto-import/pull/191, I'm seriously considering removing the typings from Handlebars 4.1 again. Would be a pity, because the minor version bump was just because of the typings. But now, it seems to be more of a breaking change than a new feature.

PR #574, opened 2019-02-09
Add repo directory key to published package.json

styfle:
This adds support for [npm RFC 10](https://github.com/npm/rfcs/blob/latest/accepted/0010-monorepo-subdirectory-declaration.md) as defined in [this PR](https://github.com/npm/rfcs/pull/19).


greysteil:
Would love to see this merged! I've implemented a [workaround](https://github.com/dependabot/dependabot-core/blob/master/npm_and_yarn/lib/dependabot/npm_and_yarn/metadata_finder.rb#L132-L137) in Dependabot in the meantime, but having DefinitelyTyped follow npm RFC 10 would be much nicer. 😄

PR #580, opened 2019-02-24
[chokidar] Add chokidar to the dependency list

skovy:
Add `chokidar` to the dependency list so that it can be included as a dependency for other types in the DefinitelyTyped repository.

It's necessary to add to this list so that the `@types/chokidar` package can be removed. 

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33359 for more context.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -27,6 +27,7 @@ axios
 bignumber.js
 broadcast-channel
 chalk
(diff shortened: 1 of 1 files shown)


sandersn:
I just pushed the whitelist changed to production. I'll re-run the CI on DefinitelyTyped/DefinitelyTyped#33359.

PR #582, opened 2019-02-28
Add 'three' to whitelist

Methuselah96:
Add `three` to the dependency list so that it can be included as a dependency for other types in the DefinitelyTyped repository.

It's necessary to add to this list so that the `@types/three` package can be removed.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33477 for more context.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -91,6 +91,7 @@ source-map
 styled-components
 sw-toolbox
 terser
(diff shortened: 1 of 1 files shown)


sandersn:
Whitelist should be up-to-date in production now. I'll restart the CI for DefinitelyTyped/DefinitelyTyped#33477.

PR #586, opened 2019-03-05
Add del to whitelist

j-oliveras:
I'm removing it from DT because new version include types.

https://github.com/sindresorhus/del/releases/tag/v4.0.0


j-oliveras:
PR on DT: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33629

PR #589, opened 2019-03-08
fix: add expect to dependency whitelist

SimenB:
It's had types since https://github.com/facebook/jest/pull/7919, released as 24.3.0 earlier today

Needed for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33705

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -40,6 +40,7 @@ egg
 electron
 eventemitter2
 eventemitter3
+expect
 fast-glob
 fastify
 flatpickr


dependenciesWhitelist.txt
```diff
@@ -40,6 +40,7 @@ egg
 electron
 eventemitter2
 eventemitter3
+expect
```
SimenB: a bunch of other jest packages also has types now, but I guess this list should only contain the packages we know we need and not everything that's theoretically possible?

sandersn: Yes, basically.

PR #594, opened 2019-03-15
Add knex to whitelist.

mastermatt:
Knex now bundles its own types and is a dep of several other type libs.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33889


mastermatt:
fyi @sandersn

sandersn:
I updated types-publisher as well. I restarted the CI build and it should now pass (or fail with different errors at least).

PR #605, opened 2019-04-16
test-runner now shells one node for each dtslint

sandersn:
It uses the new, local test-worker, which just calls cp.exec(node dtslint) for each dtslint, instead of reusing the same dtslint over and over again.

lodash and reactstrap still run out of memory, but they do that for individual calls to dtslint as well. I think multi-Typescript compiles in ExpectRule may need the same out-of-proc treatment.

Edit: No, types/lodash and types/reactstrap also run out of memory when just compiling with ts 3.4.


sandersn:
Actually, when I think about it, only lodash and react are seeing oomemory crashes on CI. lodash itself runs out of memory and reactstrap depends on react.

So I think the root cause is 3.4+ running out of memory on lodash and reactstrap. This PR isn't really needed.

sandersn:
Filed Microsoft/TypeScript#30973 to track this instead.

PR #606, opened 2019-04-19
Print performance output

sandersn:
By reading the files written by ExpectRule. If no such files exist, nothing is written in the PERFORMANCE section.

### DIFF
--- a/src/tester/test-runner.ts
+++ b/src/tester/test-runner.ts
@@ -202,7 +202,6 @@ async function doRunTests(
         return;
     }
 
-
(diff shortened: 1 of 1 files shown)


sandersn:
Next step will be to get typescript-bot to read this after it notices that CI has passed, while it is tagging PR and commenting on it.

In the meantime, I'll tell DT maintainers to ask PRs to popular packages to make sure that the type count doesn't explode.

sandersn:
OK, I'll publish it and make sure nothing explodes

PR #607, opened 2019-04-20
Add scaling crash recovery for types tester

rbuckton:
Rather than crashing the types tester when a forked process runs out of memory, the tester will now perform crash recovery using a scaling recovery mechanism:

1. By default, each forked process starts in a *crash recovery state* of `Normal`.
1. If a forked process crashes due to running out of memory and its *crash recovery state* is `Normal`, we will restart the forked process to try again:
    1. For the crashing process, its *crash recovery state* is incremented to `Retry`.
    1. If the test succeeds, the *crash recovery state* is reset to `Normal`.
1. If the forked process crashes while its *crash recovery state* is `Retry`, we will restart the forked process again with the following changes:
    1. For the crashing process, its *crash recovery state* is incremented to `RetryWithMoreMemory`.
    1. The forked process is restarted with `--max_old_space_size=4096`.
    1. If the test succeeds, the *crash recovery state* is reset to `Normal` and the forked process is restarted with the default `--max_old_space_size` settings.
1. If the forked process crashes while its *crash recovery state* is `RetryWithMoreMemory`, we will fall back to crashing the root process (existing behavior).

Whenever the *crash recovery state* of a forked process changes, a message will be written to the log indicating which package caused the crash and the current recovery state.

For example:
```
…
clean-css OK
cli-color OK
cli-progress OK
client-sessions OK
cli-spinner OK
cloneable-readable OK
cls-hooked OK
codependency OK
co-body OK

<--- Last few GCs --->

[22336:000002A80B561640]   397502 ms: Scavenge 1356.3 (1421.6) -> 1355.8 (1422.1) MB, 6.9 / 0.0 ms  (average mu = 0.150, current mu = 0.124) allocation failure
[22336:000002A80B561640]   397515 ms: Scavenge 1356.6 (1422.1) -> 1356.1 (1422.6) MB, 9.1 / 0.0 ms  (average mu = 0.150, current mu = 0.124) allocation failure
[22336:000002A80B561640]   397530 ms: Scavenge 1356.8 (1422.6) -> 1356.3 (1423.1) MB, 11.4 / 0.0 ms  (average mu = 0.150, current mu = 0.124) allocation failure


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 00000040451DC6C1]
    1: StubFrame [pc: 00000040451DDC86]
Security context: 0x009783a1e681 <JSObject>
    2: getIntersectionType(aka getIntersectionType) [000002F5E768DC79] [C:\Users\rbuckton\.dts\typescript-installs\2.9\node_modules\typescript\lib\typescript.js:~34423] [pc=0000004047C96D4A](this=0x0316a75826f1 <undefined>,types=0x00089a602c91 <JSArray[2]>,aliasSymbol=0x0316a75826f1 <undefined>,aliasTypeArguments=0x0...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF72420EEE5
 2: 00007FF7241E8CD6
 3: 00007FF7241E96E0
 4: 00007FF724650D3E
 5: 00007FF724650C6F
 6: 00007FF72459C594
 7: 00007FF724592B67
 8: 00007FF7245910DC
 9: 00007FF72459A0B7
10: 00007FF72459A136
11: 00007FF7246BF7B7
12: 00007FF7247987FA
13: 00000040451DC6C1
reactstrap Out of memory: retrying
reactstrap OK
...
```

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -4,6 +4,7 @@ import * as crypto from "crypto";
 import moment = require("moment");
 import * as os from "os";
 import * as sourceMapSupport from "source-map-support";
+import * as yargs from "yargs";
 sourceMapSupport.install();
 
 export function assertDefined<T>(x: T | undefined, message?: string | Error | undefined): T {
@@ -139,7 +140,7 @@ export function intOfString(str: string): number {
 export function sortObjectKeys<T extends { [key: string]: unknown }>(data: T): T {
     const out = {} as T; // tslint:disable-line no-object-literal-type-assertion
     for (const key of Object.keys(data).sort()) {
-        out[key] = data[key];
+        out[key as keyof T] = data[key as keyof T];
     }
     return out;
 }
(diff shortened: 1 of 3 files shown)


src/util/util.ts
```diff
@@ -349,29 +364,105 @@ export function runWithListeningChildProcesses<In>(
...
+                        case CrashRecoveryState.Normal:
+                            crashRecoveryState = CrashRecoveryState.Retry;
+                            if (handleCrash) {
+                                handleCrash(currentInput, crashRecoveryState);
+                            }
+                            restartChild();
+                            return;
+                        case CrashRecoveryState.Retry:
+                            // skip crash recovery if we're already passing a value for --max_old_space_size that is >= crashRecoveryMaxOldSpaceSize
+                            if (maxOldSpaceSize < crashRecoveryMaxOldSpaceSize) {
```
sandersn: why the secondary check? is 4096 just a hard limit right now, even if the command line specifies more?

src/util/util.ts
```diff
@@ -349,29 +364,105 @@ export function runWithListeningChildProcesses<In>(
...
+
+            const onClose = (code: number, signal: string) => {
+                if (rejected || inputIndex === inputs.length) {
+                    return;
                 }
-            });
-            child.on("close", () => { assert(rejected || inputIndex === inputs.length); });
-            child.on("error", fail);
+
+                if (crashRecovery && code === 134 && signal === null) {
```
sandersn: might comment that 134 is a magic value

src/util/util.ts
sandersn: or is there a second retry if the command line specifies <4096?

sandersn:
Based on the team room discussion we just had, could you disable RetryWithMoreMemory? It's useful for dtslint-runner, but for DT's CI, we want to require definitions to compile inside the default memory limit.

src/util/util.ts
rbuckton: If we specify `--max_old_space_size` to the root process then it wouldn't make sense to retry with "more" memory, since `fork` copies the `execArgv` of the parent to each child (i.e. each child would already be running with `--max_old_space_size` set).

src/util/util.ts
rbuckton: :+1:

rbuckton:
I can leave it in but if `crashRecoveryMaxOldSpaceSize` is set to `0` then it won't trigger it. This would allow us to turn it back on if we decide we need the capability.

PR #608, opened 2019-04-23
Add more failure logging

rbuckton:
Adds fixes and additional logging from https://github.com/DefinitelyTyped/dtslint-runner/pull/7


sandersn:
Looks like the build continues now but doesn't actually recover; it looks like all but a few packages after lodash also report oomemory.

rbuckton:
I'm looking into it.

rbuckton:
I've updated this with the latest changes from https://github.com/DefinitelyTyped/dtslint-runner/pull/9

PR #609, opened 2019-04-25
Stop using optional catch binding which is not supported in node < 10

KentoMoriwaki:
https://github.com/Microsoft/types-publisher/pull/608 added optional catch binding which is not supported in node < 10. This packages supports node >=6.10.0 According to [package.json](https://github.com/Microsoft/types-publisher/blob/master/package.json#L83), this package supports node >=6.10.0. 

I think we have two options to solve this problem
1. Stop using optional catch binding
2. Change TypeScript build target from esnext https://github.com/Microsoft/types-publisher/blob/master/tsconfig.json#L15

In this PR, I selected the former option because we need to fix quickly broken DefinitelyTyped CI pipeline. ( see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34973 )

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -538,7 +538,7 @@ export function runWithListeningChildProcesses<In>(
                     try {
                         child.removeAllListeners();
                         child.kill();
-                    } catch {
+                    } catch (e) {
                         // do nothing
                     }
                 }


KentoMoriwaki:
@rbuckton Cloud you take a look?

pluma:
Not a syntax error I think https://mariusschulz.com/blog/typescript-2-5-optional-catch-binding

pluma:
Problem seems to be that DefinitelyTyped travis uses node 8, support for the optional catch binding proposal (stage 4 now, so will likely be part of ES2019) was only added in node 10:

https://h3manth.com/new/blog/2018/whats-new-in-node-v10/

KentoMoriwaki:
@pluma Thanks for the further investigation. I updated title and description of this PR.

ExE-Boss:
This&nbsp;got&nbsp;fixed in&nbsp;https://github.com/microsoft/types-publisher/commit/9de71ee4db1979e62c8030cd7776c2899990938e.

KentoMoriwaki:
Thanks all!

PR #615, opened 2019-05-14
Add execArgv option to runWithChildProcess

andrewbranch:
Will allow me to `--prof` processes I spin up


andrewbranch:
Hang on, I need to make this fancier

andrewbranch:
Actually, I just realized this doesn't help me. In order to usefully `--prof` something, it needs to do just one thing, and these utils keep a constant number of processes alive that loop over inputs, so each process does a lot of things.

PR #621, opened 2019-06-08
added cosmiconfig to whitelist

bmatcuk:
postcss-load-config depends on an older version of cosmiconfig so I need this for my postcss-load-config definitions.

I have a PR open for that [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36053)

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -7,6 +7,7 @@
 @babel/types
 @material-ui/core
 @sentry/browser
+@types/cosmiconfig
 @types/firebase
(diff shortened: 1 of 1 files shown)


bmatcuk:
Actually, hold off on this for a moment. I've got a PR out to postcss-load-config to update the version of cosmiconfig which would make this irrelevant. If they accept that PR and release a new version in a timely manor, I can close this...

bmatcuk:
My PR to postcss-load-config was accepted, so I no longer need this. Thanks!

PR #629, opened 2019-07-01
Add bookshelf to dependenciesWhitelist.txt for use in @types/bookshelf

jwhitmarsh:
I'm going to submit a PR to update @types/bookshelf to depend on [bookshelf](https://github.com/bookshel) rather than knex. This will ensure the types align more easily.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -34,6 +34,7 @@ aws-sdk
 axe-core
 axios
 bignumber.js
+bookshelf
 broadcast-channel
 cac
 chalk


jwhitmarsh:
Hi @sandersn, are you able to review this? Is there anything I can do to help?

sandersn:
Sorry, I missed this before I went on vacation. 

Are you planning to remove bookshelf entirely? I’m not sure why you’d make bookshelf on DT depend on bookshelf itself, since it sounds like bookshelf ships its own types now. It might help to create the PR on DT to explain.

jwhitmarsh:
Does bookshelf ship its own types? I can't find any types in the project, and if I remove @types/bookshelf from my project it won't build.

I would like @types/bookshelf to depend directly on bookshelf because that way it'll track bookshelf's knex version as it's updated - currently @types/bookshelf has to be updated as an extra step, which means in userland updating to new versions of bookshelf is always delayed until @types/bookshelf is updated. 

Apologies if i've explained that poorly - I've rewritten that paragraph a few times, so hopefully it's clear!

sandersn:
OK, that *sort of* makes sense. :) I merged and updated types-publisher.

PR #631, opened 2019-07-12
Adds lit-element to the dependencies whitelist for use in @types/storybook__polymer

ryuhhnn:
I'm going to be submitting a PR for @types/storybook__polymer that requires typedefs created in `lit-element`.


sandersn:
This is now published and ready to go.

PR #635, opened 2019-07-22
Add 'express-graphql' dependencies to whitelist

IvanGoncharov:
I added TS typings to `express-graphql` and now I'm trying to remove it from DT but as I understand I need to whitelist it first:

> Error: In package.json: Dependency express-graphql not in whitelist.
If you are depending on another `@types` package, do *not* add it to a `package.json`. Path mapping should make the import work.
If this is an external library that provides typings,  please make a pull request to types-publisher adding it to `dependenciesWhitelist.txt`.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -55,6 +55,7 @@ electron-osx-sign
 eventemitter2
 eventemitter3
 expect
+express-graphql
 fast-glob
 fastify
 flatpickr


sandersn:
This is now in production.

IvanGoncharov:
@sandersn Thanks for quick merge 👍 
However, then I opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37199 to remove this library, I got:
https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/564906280#L229
```
Error: In package.json: Dependency express-graphql not in whitelist.
If you are depending on another `@types` package, do *not* add it to a `package.json`. Path mapping should make the import work.
If this is an external library that provides typings,  please make a pull request to types-publisher adding it to `dependenciesWhitelist.txt`.
```

Can you please take a look and advice on how to move forward?

sandersn:
Might be a Travis cache problem. I'll take a look.

sandersn:
I re-ran both Travis and Azure builds. Azure now fails with a different error (some dependents of express-graphql don't work with its shipped types). I cleared Travis' cache and it now gives the same error as Azure.

PR #641, opened 2019-08-09
Add xpath to whitelist

eheikes:
This is required for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37484 to use the types bundled with xpath.


sandersn:
Definitely Typed's CI should pick up the new whitelist next time it runs.

PR #642, opened 2019-08-11
Content-Encoding header is usually lowercase

bodinsamuel:
I tried downloading, with a script, the JSON file created by `npm run index` but the header content-encoding is set to `GZIP` instead of the usual `gzip`.
I'm not sure if it can be uppercase but the RFC is too much for me ahah; browsers obviously support both but some scripts/packages do not expect uppercase value.


sandersn:
Thanks!

sandersn:
This is now in production, so you'll see the fix in about half an hour if there are no publishes triggered before then.

bodinsamuel:
thanks a lot 🙏

PR #643, opened 2019-08-13
Add @vue/test-utils to whitelist

cimbul:
This will be needed by the types for `@testing-library/vue`.

(I haven't opened a DefinitelyTyped PR yet because I'm waiting on `@testing-library/dom` to be moved there first. You can see testing-library/vue-testing-library#74 if you're interested.)


sandersn:

PR #644, opened 2019-08-14
Added winston-transport as a whitelisted dependency for use in my @ty…

mhintzke13:
Simple update to the dependency whitelist so the my other PR to add @types/datadog-winston can work properly according to the documentation.


sandersn:

PR #645, opened 2019-08-15
[Fix #318] Update test-filename error to mention test directory

### LINKED ISSUE #318: Should allow multiple test files
Error [`Expected file ... to be named '...-tests.ts'`](https://github.com/Microsoft/types-publisher/blob/master/src/lib/definition-parser.ts#L158) is too restrictive.

Sometimes you need to have multiple test files.

Edit: see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/15170, it's a good use case

devrelm:
Per #318, it's not clear that multiple test files are supported by putting them in a `test/` directory.

This PR updates the error message from:

`Expected file '${file}' to be named ${expectedName}`

to:

`Expected file '${file}' to be named '${expectedName}' or to be inside a '${directoryPath}/test/' directory`


sandersn:
Fixes #318

PR #646, opened 2019-08-19
Log dependent packages in version error case.

sandersn:
For version errors, calculate and log dependent packages. Previously,
version errors were checked before the entirety of dependencies were
calculated. This gives  an iterative, annoying flavour to
fixing dependency errors:

1. Bump required version of express.
2. Get error on acl:
  Error: acl depends on express but has a lower required TypeScript version.
3. Bump required version of acl.
4. Get error on api-error-handler:
  Error: api-error-handler depends on express but has a lower required TypeScript version.
5. Bump required version of api-error-handler, etc etc.

Now an error thrown from check-require-version runs getAffectedPackages
and logs the dependents it finds. Then it re-throws the error. The above
process is now:

1. Bump required version of express.
2. Get error on acl, api-error-handler, etc etc.
3. Bump required version on all these packages.

The additional output looks like this:

Running: git rev-parse --verify master
06ddc1ccad76a7a5f766328a2fd0244bbfb06adf
Running: git diff master --name-status
M	types/express/index.d.ts
Testing 1 changed packages: express
Testing 234 dependent packages:acl,actions-on-google,api-error-handler,...

### DIFF
--- a/src/tester/test-runner.ts
+++ b/src/tester/test-runner.ts
@@ -16,7 +16,7 @@ import { npmInstallFlags } from "../util/io";
 import { consoleLogger, Logger, LoggerWithErrors, loggerWithErrors } from "../util/logging";
 import { assertDefined, exec, execAndThrowErrors, flatMap, joinPaths, logUncaughtErrors, mapIter, nAtATime, numberOfOsProcesses, runWithListeningChildProcesses, CrashRecoveryState } from "../util/util";
 
-import { getAffectedPackages, Affected, allDependencies } from "./get-affected-packages";
+import { getAffectedPackages, allDependencies } from "./get-affected-packages";
 
 const perfDir = joinPaths(os.homedir(), ".dts", "perf");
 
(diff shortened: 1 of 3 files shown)


src/tester/test-runner.ts
```diff
@@ -68,6 +68,17 @@ export default async function runTests(
     nProcesses: number,
     selection: "all" | "affected" | RegExp,
 ): Promise<void> {
+    const { changedPackages, dependentPackages, allPackages } = await getAffectedPackagesFromDiff(dt, definitelyTypedPath, selection);
```
sandersn: this change is basically an Extract Method, with an associated addition of `allPackages` to the return type.

src/tester/test.ts
```diff
@@ -22,6 +22,15 @@ async function main(options: TesterOptions, nProcesses: number, all: boolean): P
     const log = loggerWithErrors()[0];
     const dt = await getDefinitelyTyped(options, log);
     await parseDefinitions(dt, { nProcesses, definitelyTypedPath: options.definitelyTypedPath }, log);
-    await checkParseResults(/*includeNpmChecks*/false, dt, options, new UncachedNpmInfoClient());
+    try {
+        await checkParseResults(/*includeNpmChecks*/false, dt, options, new UncachedNpmInfoClient());
+    }
+    catch (e) {
+        if (!all) {
```
sandersn: this is the actual change. I decided to skip printing for "all" because presumably the list of dependent packages will be every package that depends on another.

src/tester/test.ts
```diff
@@ -22,6 +22,15 @@ async function main(options: TesterOptions, nProcesses: number, all: boolean): P
     const log = loggerWithErrors()[0];
     const dt = await getDefinitelyTyped(options, log);
     await parseDefinitions(dt, { nProcesses, definitelyTypedPath: options.definitelyTypedPath }, log);
-    await checkParseResults(/*includeNpmChecks*/false, dt, options, new UncachedNpmInfoClient());
+    try {
+        await checkParseResults(/*includeNpmChecks*/false, dt, options, new UncachedNpmInfoClient());
+    }
+    catch (e) {
+        if (!all) {
+            (await getAffectedPackagesFromDiff(dt, options.definitelyTypedPath, "affected")).dependentPackages.map(t => t.desc);
```
andrewbranch: Did you mean to do something with the result of this expression, or do you not need the `.dependentPackages.map(...)`, or am I missing something?

sandersn: I meant to use it until I noticed that getAffectedPackagesFromDiff already logged the information. Removed!

PR #649, opened 2019-08-22
add 'graphql' to whitelist

IvanGoncharov:
Please see: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37830


sandersn:

PR #653, opened 2019-09-06
make error message for @types dependencies clearer

dbartholomae:
While working on my [first PR to DefinitivelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37908), I had some trouble understanding how to handle internal dependencies between types, specifically if the dependency is namespaced. To spare others the confusion I have added more info to the error message.


dbartholomae:
I'm working on Windows, so most likely a line-ending thing. I'll take another look :)

dbartholomae:
I've tried both LF and CRLF, but it doesn't seem to work :-/
Not sure how to fix the whitespace.

PR #656, opened 2019-09-16
Add @emotion/styled to whitelist

iamkd:
This is required by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38411 to use the StyledComponent definition from Emotion.


sandersn:

PR #657, opened 2019-09-17
Add @emotion/core

thebuilder:
This is needed for Rebass typings: 
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38433#issuecomment-532138518


sandersn:

PR #659, opened 2019-09-20
Move deploys to run on CI

orta:
You as a maintainer needs to make a commit, and push a tag and CI will deploy


orta:
Ah, yeah, then in context this doesn't make sense at all 👍

PR #662, opened 2019-10-09
Whitelist @storybook/addons

MichaelDeBoey:
DefinitelyTyped/DefinitelyTyped#38961 needs this.

CC @sandersn @RyanCavanaugh @orta


sandersn:

PR #663, opened 2019-10-09
Allow suppressing messages back to github PRs via a commit message

orta:
Supports crafting a specially named commit which surpasses the GitHub message after publishing.

### DIFF
--- a/src/publish-packages.ts
+++ b/src/publish-packages.ts
@@ -52,12 +52,15 @@ export default async function publishPackages(
             fetcher) as Array<{
             sha: string,
             commit: {
+                message: string,
                 author: {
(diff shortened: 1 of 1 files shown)


weswigham:
Why, though? On DT we should _all_ _always_ be using a squash and merge, so if a user includes this in _any_ of their commit messages (which we don't police), the bot won't post back to the PR and we'll be very confused as to why.

orta:
This was in preparation for a one-off commit which ran prettier, and wouldn't force a re-deploy. All normal flow in DT wouldn't be able to trigger it on purpose 👍  (in theory this could be removed if we wanted now)

PR #665, opened 2019-10-15
Update whitelist

ViceIce:
[knockout](https://github.com/knockout/knockout/blob/master/build/types/knockout.d.ts) has now it's own types

I'm working on https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37490


sandersn:

PR #668, opened 2019-10-17
Add rrule to whitelist

nickbclifford:
Needed for DefinitelyTyped/DefinitelyTyped#39198.


sandersn:

PR #669, opened 2019-10-17
Add query-string package to whitelist

jsamr:
See PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39197
proper-url-join depends on [query-string](https://www.npmjs.com/package/query-string)


sandersn:

PR #671, opened 2019-10-22
Added adal-node to whitelist

armanio123:
Adds adal-node to dependenciesWhitelist.txt file.


sandersn:

PR #672, opened 2019-10-29
Add actions-on-google to whitelist

nightillusions:
Added this dependency for an new typing in DefinitelyTyped


sandersn:

PR #686, opened 2019-11-01
add sequelize to dependencies whitelist

a-tarasyuk:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39543

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -145,6 +145,7 @@ should
 smooth-scrollbar
 scroll-into-view-if-needed
 source-map
+sequelize
 styled-components
 sw-toolbox
 swagger-parser


papb:
What does this mean?

sandersn:
@papb Usually, it's needed when a popular package starts providing its own types. Somebody (@a-tarasyuk in this case) removes the types from Definitely Typed and has to update all the other @types packages that depend on it to point to the actual package.

To make sure that @types packages don't depend on possibly-malicious packages, we require a human to update a whitelist and publish a new version of the CI test code.

@a-tarasyuk I read the PR comments and it looks like the owners want to keep the types around for v3 and v4. @papb does that agree with your understanding? I won't merge this until DefinitelyTyped/DefinitelyTyped#39543 comes to a consensus.

a-tarasyuk:
@sandersn Oke. Should I close both PR? Or is it OK to keep them open?

sandersn:
I would close both if we're not going to take them. (The bot will close the DT one if you leave it alone for a while.)

a-tarasyuk:
Oke. will close

papb:
@sandersn Thanks for the explanation

> I read the PR comments and it looks like the owners want to keep the types around for v3 and v4. @papb does that agree with your understanding?

Yes, I agree, I think the types should be kept around for v3 a v4. Thanks!

PR #694, opened 2019-11-01
Dual-publish Definitely Typed packages to github package registry

sandersn:
This PR is the next step in using the github package registry to support Automatic Type Acquisition. It generates two packages side-by-side on disk and calls `npm publish` on each. But the packages differ only in that the github package.json has a `publishConfig` entry that specifies the github package registry. This is the same strategy as the dual-publish of types-registry.

I plan to let this run for a few weeks until the package registry hits GA, then switch to publishing to the `@types` scope on the Types org.

Some notes:

1. I switched from "npm" | "github" to an enum, including in old code.
2. I still need to write tests of package removal and actual publishing. I'm not sure how annoying mocking the npm client will be, however. Since I'm on DT next week, watching closely all week after the merge will be worth more than unit tests. The unit tests are really for future changes when I've forgotten how all this works again.
3. The temporary target is `@testtypepublishing`, but the eventual target should be `@types` in order to provide a complete mirror of the `@types` scope on npm.

### DIFF
--- a/src/generate-packages.test.ts
+++ b/src/generate-packages.test.ts
@@ -1,6 +1,6 @@
 import { createPackageJSON, createReadme, getLicenseFileText } from "./generate-packages";
 import { Registry } from "./lib/common";
-import { TypingsData, TypingsDataRaw, License, AllPackages, readNotNeededPackages } from "./lib/packages";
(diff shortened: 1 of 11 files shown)


sandersn:
The package.json tests read the file system :sweat: so I've got *more* things to mock.

orta:
You got this!

PR #697, opened 2019-11-06
add @types/expect to dependencies whitelist

a-tarasyuk:
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39618


sandersn:

PR #700, opened 2019-11-07
remove ink from Whitelist in favor of @types/ink

a-tarasyuk:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39595


sandersn:
Same as the highcharts question -- why was ink in the whitelist if nothing actually depended on it?

I checked, and nothing currently does. I guess some package had previously depended on it?

PR #706, opened 2019-11-08
remove next-redux-wrapper from Whitelist in favor of @types/next-redux-wrapper

a-tarasyuk:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39478


sandersn:

PR #709, opened 2019-11-18
Rename UNUSED_FILES to OTHER_FILES

sandersn:
Also skip blank lines from the file


sandersn:
@andrewbranch I put you on the review for this but it's so simple that I'll just merge it after CI passes. (Not that there are any tests for UNUSED_FILES.txt right now...)

Issue #708, opened 2019-11-18
index.d.ts should be the only d.ts file allowed in tsconfig.json


sandersn:
Currently, tsconfig.json for each package contains all the test files and all the type files. However, for the test files, this doesn't reflect how users will reference the type files as installed in `@types/*`. This matters when a package has a global augmentation:

```ts
// @filename: globals.d.ts
declare global {
   function glo(n: number): void
}
// @filename: index.d.ts
export const x = 1

// @filename: package-tests.ts
import { x } from 'package'
// @filename: global-tests.ts
glo(1)
```

When `@types/package` is actually installed, the last file, `global-tests.ts`, is incorrectly able to resolve `glo` without having 

```
/// <reference types="package/globals" />
```

At the top. That's because the in-DT tsconfig currently looks like this:

```json
...
"files": [
  "index.d.ts",
  "globals.d.ts",
  "package-tests.ts",
  "global-tests.ts"
]
```

Only `index.d.ts` should be allowed of the types files:

```json
...
"files": [
  "index.d.ts",
  "package-tests.ts",
  "global-tests.ts"
]
```

Everything else should be referenced either through `index.d.ts` or through an explicit reference in the tests. That's because users shouldn't have references in their `"files"` like `node_modules/@types/package/global-tests.ts`.

Fixing this will require a couple of changes:

1. dtslint will need a rule that checks tsconfig and disallows all d.ts except `index.d.ts`.
2. types-publisher will need to follow dependencies in tests too. Right now it doesn't.

d.ts files that aren't tested and aren't used need to be added to `UNUSED_FILES.txt`. Previously they could just be added to `"files"` in tsconfig. Because UNUSED_FILES.txt will not include files that *might* be used, but happen to have no tests, I'm going to rename it to OTHER_FILES.txt.

So, this bug will have five associated PRs:

- [x] Types-publisher: Rename UNUSED_FILES.txt to OTHER_FILES.txt (#709)
- [x] Rename 6 UNUSED_FILES.txt to OTHER_FILES.txt in existing DT packages. (DefinitelyTyped/DefinitelyTyped#40479)
- [x] Types-publisher: start following references in tests for calculating unused files. (#710)
- [ ] Drop already-referenced files from about 250 tsconfig.jsons. (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40676)
[...]

PR #710, opened 2019-11-21
Follow test references to calculate unused files

sandersn:
To make sure that I didn't break anything, I dumped the list of dependencies generated during parsing. In `getTypingInfo`, I added the following code:

```ts
    console.log(`"${packageName}": {`)
    let first = true
    for (const k in res) {
        if (first)
            first = false
        else
            console.log(",")
        console.log(`  "${k}": ${JSON.stringify(res[k].dependencies.map(d => d.name))}`)
    }
    console.log("},")
```

I put the same code in the production version of types-publisher. This produced (1) almost-legal JSON (2) a diffable file for both. I diffed the files and found 57 differences:

* 30: order differences from sorting or not. 
* 2: incorrect `types="x/v7"` references that are now corrected to just `types="x"` by the additional call to `rootName`. I added a test to show that we detect and prevent such references.
* 2: incorrect references "chrome/chromecast" from `types="chrome/chromecast"` are corrected to just "chrome" by the additional call to `rootName`.
* 1: incorrect reference to "@ember/services", from @types/ember__services itself.
* 22: missing references from untested files. I added entries to OTHER_FILES.txt to add these missing references.

There were lots of unused files that I automatically added OTHER_FILES.txt entries for. That's the next step in the process I wrote up at #708.

Note: I *also* need to figure out how to fix the 2 references to `types="x/v7"` which the tests relied on. I think I'll have to add them to the whitelist. =[
Edit: I just updated the tests to newer versions of dependencies: DefinitelyTyped/DefinitelyTyped#40575

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -156,31 +156,23 @@ async function getTypingDataForSingleTypesVersion(
 ): Promise<TypingDataFromIndividualTypeScriptVersion> {
     const tsconfig = await fs.readJson("tsconfig.json") as TsConfig; // tslint:disable-line await-promise (tslint bug)
     await checkFilesFromTsConfig(packageName, tsconfig, fs.debugPath());
-    const [typeFiles, testFiles] = await allReferencedFiles(tsconfig.files!, fs, packageName, packageDirectory);
-    const allUsedFiles = new Set([...typeFiles.keys(), ...testFiles.keys(), "tsconfig.json", "tslint.json"]);
-    const otherFiles = await checkAllFilesUsed(ls, allUsedFiles, fs);
+    const { types, tests } = await allReferencedFiles(tsconfig.files!, fs, packageName, packageDirectory);
+    const usedFiles = new Set([...types.keys(), ...tests.keys(), "tsconfig.json", "tslint.json"]);
+    const otherFiles = await checkAllFilesUsed(ls, usedFiles, packageName, fs);
     for (const untestedTypeFile of filter(otherFiles, name => name.endsWith('.d.ts'))) {
-        typeFiles.set(untestedTypeFile, createSourceFile(untestedTypeFile, await fs.readFile(untestedTypeFile)));
+        // add d.ts files from OTHER_FILES.txt in order get their dependencies
+        types.set(untestedTypeFile, createSourceFile(untestedTypeFile, await fs.readFile(untestedTypeFile)));
     }
 
-    const { dependencies: dependenciesWithDeclaredModules, globals, declaredModules } = await getModuleInfo(packageName, typeFiles);
+    const { dependencies: dependenciesWithDeclaredModules, globals, declaredModules } = await getModuleInfo(packageName, types);
     const declaredModulesSet = new Set(declaredModules);
     // Don't count an import of "x" as a dependency if we saw `declare module "x"` somewhere.
     const dependenciesSet = new Set(filter(dependenciesWithDeclaredModules, m => !declaredModulesSet.has(m)));
-    const testDependencies = Array.from(filter(await getTestDependencies(packageName, testFiles.keys(), dependenciesSet, fs), m => !declaredModulesSet.has(m)));
+    const testDependencies = Array.from(filter(await getTestDependencies(packageName, tests.keys(), dependenciesSet, fs), m => !declaredModulesSet.has(m)));
 
     const { dependencies, pathMappings } = await calculateDependencies(packageName, tsconfig, dependenciesSet, oldMajorVersion);
-
-
-    // Double-check that no windows "\\" broke in.
-    for (const fileName of allUsedFiles) {
-        if (hasWindowsSlashes(fileName)) {
-            throw new Error(`In ${packageName}: windows slash detected in ${fileName}`);
-        }
-    }
-
     const tsconfigPathsForHash = JSON.stringify(tsconfig.compilerOptions.paths);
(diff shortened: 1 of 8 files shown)


src/lib/definition-parser.ts
```diff
@@ -155,28 +155,25 @@ async function getTypingDataForSingleTypesVersion(
     oldMajorVersion: number | undefined,
 ): Promise<TypingDataFromIndividualTypeScriptVersion> {
     const tsconfig = await fs.readJson("tsconfig.json") as TsConfig; // tslint:disable-line await-promise (tslint bug)
```
sandersn: allReferencedFiles finds all the files that are referenced transitively from the files listed in tsconfig. Then we manually reference d.ts files from OTHER_FILES.txt as references.

Next, the external dependencies are calculated. This part doesn't change too much, I just moved `checkAllFilesUsed`'s call earlier and the no-windows-\ check earlier, into `checkAllFilesUsed`.

src/lib/module-info.ts
```diff
@@ -83,7 +78,7 @@ export default async function getModuleInfo(
         }
     }
 
-    return { declFiles: sort(all.keys()), dependencies, declaredModules, globals: sort(globals) };
```
sandersn: I moved the `sort(all.keys())` out of this function since it didn't particularly belong.

src/lib/module-info.ts
```diff
@@ -7,16 +7,15 @@ import { hasWindowsSlashes, joinPaths, normalizeSlashes, sort } from "../util/ut
...
-    const all = await allReferencedFiles(allEntryFilenames, fs, packageDirectory);
+export async function getModuleInfo(packageName: string, all: Map<string, ts.SourceFile>): Promise<ModuleInfo> {
 
     const dependencies = new Set<string>();
     const declaredModules: string[] = [];
     const globals = new Set<string>();
 
-    function addDependency(dependency: string): void {
+    function addDependency(ref: string): void {
+        if (ref.startsWith(".")) return;
```
sandersn: this now happens for all calls to addDependency, which fixes a couple of bogus dependencies I saw on DT.

src/lib/module-info.ts
```diff
@@ -123,15 +116,24 @@ function properModuleName(folderName: string, fileName: string): string {
...
     let slash = importText.indexOf("/");
     // Root of `@foo/bar/baz` is `@foo/bar`
     if (importText.startsWith("@")) {
         // Use second "/"
         slash = importText.indexOf("/", slash + 1);
     }
-    return slash === -1 ? importText : importText.slice(0, slash);
+    const root = importText.slice(0, slash);
+    const postImport = importText.slice(slash + 1);
+    if (slash > -1 && postImport.match(/v\d+$/) && !typeFiles.has(postImport + ".d.ts")) {
```
sandersn: this error is new, and required a couple of packages on DT to be fixed.

src/lib/module-info.ts
```diff
@@ -140,9 +142,14 @@ function withoutExtension(str: string, ext: string): string {
 }
 
 /** Returns a map from filename (path relative to `directory`) to the SourceFile we parsed for it. */
-async function allReferencedFiles(entryFilenames: ReadonlyArray<string>, fs: FS, baseDirectory: string): Promise<Map<string, ts.SourceFile>> {
+export async function allReferencedFiles(
```
sandersn: allReferencedFiles now scans through .d.ts AND .ts files, instead of only looking at d.ts files. It uses extension d.ts-vs-ts to decide whether the referenced file is a type file or a test file.

This is not precisely equivalent to the old algorithm but it produces the same results on DT (so far at least).

src/lib/module-info.ts
```diff
@@ -292,45 +326,43 @@ function assertNoWindowsSlashes(packageName: string, fileName: string): string {
...
             }
-
             testDependencies.add(referencedPackage);
         }
-
         for (const imported of imports(sourceFile)) {
-            if (!imported.startsWith(".") && !dependencies.has(imported) && imported !== pkgName) {
-                testDependencies.add(imported);
+            if (!imported.startsWith(".")) {
+                const dep = rootName(imported, typeFiles);
```
sandersn: this call to rootName is new, and caught several mistakes on DT.

src/mocks.ts
```diff
@@ -10,6 +10,131 @@ export function createMockDT() {
...
+export const inane: true | false;
+`);
+    boring.set("untested.d.ts", `
+import { help } from 'manual';
+export const fungible: false;
+`);
+    boring.set("boring-tests.ts", `
+import { superstor } from "super-big-fun-hus";
+import { drills } from "boring";
+import { hovercars } from "boring/secondary";
```
sandersn: need to figure out how to work a cybertruck reference in here somewhere

src/lib/definition-parser.ts
```diff
@@ -261,7 +250,7 @@ interface TsConfig {
     compilerOptions: ts.CompilerOptions;
 }
 
-/** In addition to dependencies found oun source code, also get dependencies from tsconfig. */
+/** In addition to dependencies found in source code, also get dependencies from tsconfig. */
```
andrewbranch: 😐

sandersn:
I'll merge and ship once I have the related DT PR(s) ready to go.

sandersn:
I missed a last minute change, which I just pushed to master: c169523

src/lib/module-info.ts
sandersn: Much later: This should be legal if the import is from the same package, eg "foo/v3" should be legal inside "foo".
Not sure how to detect this yet.

PR #711, opened 2019-11-21
Add @types/gulp to whitelist

sandersn:
-Sigh- This is needed because gulp-connect still depends on gulp 3.


sandersn:
New plan, just make update gulp 4 examples for gulp-connect, since it's largely abandoned as a project anyway.

PR #713, opened 2019-11-27
Make filesystem interface synchronous

sandersn:
In definition-parser and module-info, this provides about a 25% speedup on my machine. I expect gains to be smaller on slower machines since the time taken to run should be relatively larger than the overhead of async.

The overall ability for parallelism remains, implemented by forking separate processes and having them communicate. This just removes fine-grained parallelism via asynchrony during filesystem operations.

It also makes debugging module-info much easier, and the order of parsing deterministic. And there's less code overall.

The only place I would expect to be slower is in generate-packages, when copying files from Definitely Typed to the package just before publishing. But this only runs 5 times per package on average and a few hundred times in the biggest packages. So I don't think making the read synchronous will kill publishing performance, which is a background process anyway.


sandersn:
@andrewbranch You don't necessarily need to sign off on this since it's a type-directed change, but I thought you might be interested. I'm going to test ATA for a while and then come back to merge this and ship it.

sandersn:
Update: On CI, where this really matters, it looks like parsing went from about 30 seconds to about 20 seconds!

sandersn:
In types-publisher, which is not otherwise parallelised, parsing went from almost 140 seconds to about 130 seconds.

PR #715, opened 2019-12-09
Remove parallelism from npm install

sandersn:
I'm not sure this will completely prevent `npm install` errors in @types/node PRs, but it should help.


sandersn:
Local testing shows no npm install errors after editing node, for what it's worth.

sandersn:
It's now in the production branch, so Definitely Typed's CI will get the new version next time it runs.

PR #717, opened 2019-12-13
Support majors as minors in folders

orta:
I ended up an hour or two deep into my first shot at this by trying to support "x.y" versions. This... became a massive refactor which I became less confident in as I dug deeper.

The tricky thing with supporting x.y versions internally is that it _has_ to be mapped to a string otherwise you lose `0.60`, or `0.0` in the conversion to a number. Now suddenly a lot of places are going to move to validating both minor and major, and there's the number coercion for comparisons etc. It got a bit rough.

This instead  does not raise when using a minor version as a major when the initial version is 0. Allowing someone with `0.61` to set a folder as `v61` and it to be 🆗 

My setup for some reason isn't letting me do a full parse on a clean compile, so I'm hoping CI will validate it here

re: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40923

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,19 @@ branches:
 
 install:
   - npm ci
+  - git clone https://github.com/DefinitelyTyped/DefinitelyTyped ../DefinitelyTyped
+
+script:
+  - npm test
+  
+  - npm run build
+  - npm run clean
+  - npm run parse
+  - npm run check
+  - npm run calculate-versions
+  - npm run generate
+  - npm run index
+
 
 git:
   depth: 1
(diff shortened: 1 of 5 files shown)


acoates-ms:
I"m not sure that it makes sense to allow v61 as a folder name for 0.61.  What happens when you have 0.2 version with v2 as the folder name, and you then want to release 2.0?  I'd rather force people to be more consistent? -- Unless this is needed for some existing packages?

acoates-ms:
@orta - Ping.  Any idea when this will happen?

orta:
I'm closing this PR, I'm on vacation, but I'll review #723

PR #718, opened 2019-12-13
Make CI slow by validating against master definitely typed

orta:
This clones DT into a sibling folder, then runs through everything up till deployment on CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,19 @@ branches:
 
 install:
   - npm ci
+  - git clone https://github.com/DefinitelyTyped/DefinitelyTyped ../DefinitelyTyped
+
+script:
+  - npm test
+  
+  - npm run build
+  - npm run clean
+  - npm run parse
+  - npm run check
+  - npm run calculate-versions
+  - npm run generate
+  - npm run index
+
 
 git:
   depth: 1


.travis.yml
```diff
@@ -11,6 +15,14 @@ branches:
...
 install:
   - npm ci
+  - git clone https://github.com/DefinitelyTyped/DefinitelyTyped ../DefinitelyTyped
+
+script:
+  - npm tes
+  - npm run build
+  - npm run clean
+  - npm run parse
+  - npm run check
```
sandersn: 1. This isn't very valuable, it mostly just tells you what `@types` packages are superceded by in-package types.
2. On the other hand, we shouldn't *never* run it.
3. On the third hand, it'll make CI slow.
4. Is CI never not slow?
3. This is probably the right place to run it.

.travis.yml
```diff
@@ -11,6 +15,14 @@ branches:
...
```
sandersn: probably should `npm run calculate-versions` and `npm run generate` and `npm run index`.

Note that some of these tasks hit npm (parse, index and maybe generate?)

.travis.yml
```diff
@@ -11,6 +15,14 @@ branches:
 
 install:
   - npm ci
+  - git clone https://github.com/DefinitelyTyped/DefinitelyTyped ../DefinitelyTyped
+
+script:
+  - npm tes
+  - npm run build
+  - npm run clean
+  - npm run parse
```
sandersn: This relies on a cache to avoid querying npm for everything. It is NOT the most reliable cache, and I don't understand travis' caching strategy. You should test to see whether it's running `npm info` on every `@types` package for every CI run. If it's *not*, then you need to make sure that it's correctly running `npm info` on packages that were recently updated in DT.

PR #720, opened 2019-12-17
Move github publishing to @types

sandersn:
1. Switch from @testtypepublishing to @types. This simplifies some NPM-vs-Github code because the package name is the same.
2. On Github. split @definitelytyped/types-registry publication from @types/* publication. They are two separate publishes to two orgs that need two tokens.

TODO: This PR requires a new token for @types registration and changes the name for the @definitelytyped token. These two are *different* from the token used to access the Github API for posting comments to PRs.

### DIFF
--- a/src/generate-packages.test.ts
+++ b/src/generate-packages.test.ts
@@ -50,23 +50,19 @@ testo({
     },
     basicReadme() {
         const typing = new TypingsData(createRawPackage(License.Apache20), /*isLatest*/ true);
-        expect(createReadme(typing, Registry.NPM)).toEqual(expect.stringContaining("This package contains type definitions for"));
-    },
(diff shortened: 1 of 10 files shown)


sandersn:
Welp, never mind, publish tokens are not per-repo, so I don't need a new one after all.
This cuts the code size in half, although it *seems* less secure. I'm not an expert, though, and non-experts often confuse complexity for quality, so I could be wrong.

I'll update this PR.

sandersn:
It's working! It's working! Here's the first package, published from my local machine: https://github.com/types/_definitelytypedmirror/packages/82311

sandersn:
And here's the first package published by types-publisher: https://github.com/types/_definitelytypedmirror/packages/82318

PR #721, opened 2019-12-24
[tslint] Make green and add to CI

alloy:
Most of this shouldn’t be controversial, I tried to follow what seemed to be current style in use, although there’s one place where logic actually changed which I’ll comment on inline.

I’ve also added linting to the `test` script. If there’s a wish to do so, I can separate that and add a `ci` script that runs both.

(In the future using prettier would make most of these style issues go away and probably ease contributions.)


alloy:
cc @orta

src/generate-packages.ts
```diff
@@ -43,7 +42,7 @@ export default async function generatePackages(dt: FS, allPackages: AllPackages,
         log(` * ${pkg.libraryName}`);
     }
     log("## Generating deprecated packages");
-    withNpmCache(new UncachedNpmInfoClient(), async client => {
+    await withNpmCache(new UncachedNpmInfoClient(), async client => {
```
alloy: The linter was complaining that there was a unhandled promise here, which appears to be true, but perhaps this was intentional to speed-up the process?

src/publish-packages.ts
```diff
@@ -134,20 +137,19 @@ export default async function publishPackages(
         }
     }
 
-    withNpmCache(new UncachedNpmInfoClient(), async infoClient => {
+    await withNpmCache(new UncachedNpmInfoClient(), async infoClient => {
```
alloy: Ditto

alloy:
Btw, I’ll let a full build run to ensure nothing broke that’s not covered by tests.

alloy:
I think it’s always a tad annoying regardless of order, to be honest. At least in this order you are able to pass on options to `jest`, which is something I at least tend to do more often. E.g. `nom run test -- --watch`.

PR #725, opened 2020-01-08
fix: Format contributors links properly

Maxim-Mazurok:
In #698 links in README on NPM were clickable, but [now](https://www.npmjs.com/package/@types/gapi.client.sheets#credits) they are just plain text.

This PR aims to do this after markdown transformation: 
```html
<a href="https://github.com/me">My Self</a>
```

**Before:**
Screenshot from 2020-01-08 17-43-46
```markdown
These definitions were written by Maxim Mazurok (https://github.com/Maxim-Mazurok).
```

**After:**
Screenshot from 2020-01-08 17-44-21
```markdown
These definitions were written by [Maxim Mazurok](https://github.com/Maxim-Mazurok).
```

So that no matter how npmjs treats plain-text links, now link will always be clickable, also consuming less space and being more readable.

### DIFF
--- a/src/generate-packages.ts
+++ b/src/generate-packages.ts
@@ -207,7 +207,7 @@ export function createReadme(typing: TypingsData): string {
     lines.push("");
 
     lines.push("# Credits");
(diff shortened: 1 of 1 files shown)


orta:
Perfect, thanks!

PR #727, opened 2020-01-14
Add @emotion/serialize to whitelist

callumacrae:
Required for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41567

Proof: https://github.com/emotion-js/emotion/tree/master/packages/serialize/types


sandersn:

PR #729, opened 2020-01-14
Fixed assert on platforms with different sort rules (#728)

### LINKED ISSUE #728: Tests do not run reliably on Windows
The underlying problem appears to be assuming the output of `readdirSync()` is sorted according to ASCII order, so that `assertSorted()` throws in `DiskFS.prototype.readdir` (in `get-definitely-typed.ts`). The specific case is `babel__code-frame` vs. `babelify`.

Adding a `sort()` fixes the issue. I'll be submitting a PR, just creating this issue for tracking purposes.

doug-a-brunner:
Fixes #728.

### DIFF
--- a/src/get-definitely-typed.ts
+++ b/src/get-definitely-typed.ts
@@ -225,7 +225,7 @@ class DiskFS implements FS {
     }
 
     readdir(dirPath?: string): ReadonlyArray<string> {
-        return assertSorted(readdirSync(this.getPath(dirPath))).filter(name => name !== ".DS_STORE");
+        return assertSorted(readdirSync(this.getPath(dirPath)).sort()).filter(name => name !== ".DS_STORE");
     }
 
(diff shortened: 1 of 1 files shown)


src/get-definitely-typed.ts
```diff
@@ -225,7 +225,7 @@ class DiskFS implements FS {
     }
 
     readdir(dirPath?: string): ReadonlyArray<string> {
-        return assertSorted(readdirSync(this.getPath(dirPath))).filter(name => name !== ".DS_STORE");
+        return assertSorted(readdirSync(this.getPath(dirPath)).sort()).filter(name => name !== ".DS_STORE");
```
sandersn: Can you get rid of the `assertSorted` call? Now, it's just testing that `sort` works.

doug-a-brunner: Done.

PR #733, opened 2020-01-20
Add `ethers` to whitelist

cag:
Will enable https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41744


sandersn:

PR #735, opened 2020-02-02
feat(connect-mongo): add connect-mongo to white list

peterblazejewicz:
As of version 3.1.1 the connect-mongo ships its own definition file.
Adding to white list will enable using 'connect-mongo' as external
dependency in the existing definitions like passport.socketio.

DefinitelyTyped/DefinitelyTyped#42047

Thanks!


sandersn:

PR #745, opened 2020-02-19
Update dependenciesWhitelist.txt

PrasadKasireddy:
Update the Dependencies list

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -23,6 +23,7 @@
 @storybook/addons
 @storybook/react
 @types/base-x
+@types/browserify
 @types/expect
 @types/firebase
 @types/highcharts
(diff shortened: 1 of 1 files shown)


PrasadKasireddy:
> exemptions

I can see these are dependency with other packages. I assume that these are missed.
Please refer below URL. 
@types/browserify - http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Please review and correct me if am wrong.

sandersn:
Can you point to a PR on DT that goes with this PR? Typically, to use other `@types` packages inside of DT, you just import normally eg `import aframe = require('aframe')` and it will resolve to `~/DefinitelyTyped/types/aframe/index.d.ts`

PrasadKasireddy:
> DefinitelyTyped/types/aframe/index.d.ts

Thanks for your clarification

PR #748, opened 2020-02-27
Log suggestions

gabritto:
types-publisher will now collect suggestions from the [files that dtslint wrote](https://github.com/microsoft/dtslint/pull/276), and print them (a single line per package).
types-publisher runs dtslint for each DT package in parallel, but since the suggestion files are per-package, there should be no data races.


gabritto:
CI seems to be failing due to lint errors that were already present.

PR #750, opened 2020-03-02
Fix lint errors

gabritto:
Most fixes are removing an unnecessary `else` or calling `.toString()`.
I've added `strict-string-expressions` exemptions where the fix wasn't obvious, and a `no-async-without-await` exemption as well.
I've also changed the TSLint arguments for `variable-name` to allow `catch (_) { ... }` (i.e. leading underscores) to be valid and `strict-comparisons` to allow string comparisons and object comparisons, as there were lots of occurrences of those patterns.

### DIFF
--- a/src/lib/packages.ts
+++ b/src/lib/packages.ts
@@ -247,7 +247,7 @@ export class NotNeededPackage extends PackageBase {
         super(raw);
         this.sourceRepoURL = raw.sourceRepoURL;
 
-        for (const key in raw) {
+        for (const key of Object.keys(raw)) {
             if (!["libraryName", "typingsPackageName", "sourceRepoURL", "asOfVersion"].includes(key)) {
                 throw new Error(`Unexpected key in not-needed package: ${key}`);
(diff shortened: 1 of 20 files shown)


src/lib/packages.ts
```diff
@@ -456,7 +455,7 @@ export class TypingsVersions {
 
         this.map = new Map(this.versions.map(version => {
             const dataKey = versionMappings.get(version)!;
-            return [version, new TypingsData(data[dataKey], version === this.versions[0])];
+            return [version, new TypingsData(data[dataKey], version.equals(this.versions[0]))];
```
sandersn: what is the type of `version`? SemVer or string? It looks like this caught a real bug here.

gabritto: TS says the type is `Semver`. From what I understandd, this was working because luckily we were only comparing `Semver` objects that were part of the same array, so `===` and `.equals` have the same effect in this particular situation.

PR #751, opened 2020-03-04
Updating whitelist to include htmlparser2

daviduzumeri:
For enabling https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42841


sandersn:

PR #754, opened 2020-03-13
Add `jimp` to the deps whitelist.

peterblazejewicz:
The `jimp` ships with its own type definitions. The definition files for
the consumers of that package need to install this via dependency list in
package.json.

https://github.com/oliver-moran/jimp

Thanks!


sandersn:

peterblazejewicz:
Thanks!

PR #757, opened 2020-03-29
Added 'node-php-password'

mikilian:
I want to provide type support for [node-php-password](https://github.com/thomas-alrek/node-php-password).

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -132,6 +132,7 @@ moment-range
 mqtt
 next
 nock
+node-php-password
 normalize-url
 objection
 opentracing


sandersn:
The list you edited allows other `@types` packages to refer to node-php-password, if node-php-password provides its own types. But it doesn't, so you need to send a PR to Definitely Typed instead, in order to create `@types/node-php-password`.

sandersn (changes requested):
Seems like a PR to Definitely Typed is the correct thing here.

andrewbranch:
👋 @michaelkilian we’re preparing to archive this repo since it has [moved](https://github.com/microsoft/types-publisher#-this-repo-is-in-the-process-of-moving), and since it seems like this PR is not currently mergeable, I’m going to close it. Thanks!

PR #758, opened 2020-03-29
Whitelist `typescript-tuple` library

Igorbek:
This library along with `ts-toolbelt` (which is already in the list) provides very convenient helpers to operate on tuple types.
This is very common for typings to operate on tuple types, and for authors, it is easier to use well-designed solutions.
This particular library works with tuples a little better than `ts-toolbelt` and provides more helper types.

npm: [`typescript-tuple`](https://www.npmjs.com/package/typescript-tuple)


sandersn:

PR #759, opened 2020-03-30
Whitelist got

Richienb:
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43453


sandersn:

PR #763, opened 2020-04-06
whitelist acorn

djcsdy:
acorn distributes its own types since v6.0.0.

See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43670

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -45,6 +45,7 @@
 abort-controller
 actions-on-google
 activex-helpers
+acorn
 adal-node
 ajv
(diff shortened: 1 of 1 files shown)


djcsdy:
Turns out the types distributed with acorn are absolutely useless. Don't merge this. Sorry for the noise.

sandersn:
Good to know! Interestingly, we have acorn in our user tests, so we would have been testing the useless types, except that the test is set up to test the distributed javascript instead. I think I added the test during a push to improve checkJS and wanted to compile large, popular JS projects.

PR #764, opened 2020-04-07
Whitelist @types/got

Richienb:
See https://travis-ci.org/github/DefinitelyTyped/DefinitelyTyped/builds/669145664#L234. Related to: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43453


sandersn:
This is unneeded since we don't currently have two `@types` packages where one depends on `got` (current version) and another depends on `@types/got` (old version preserved on DT for shipping fixes).

PR #766, opened 2020-04-20
Add glaze and treat to dependenciesWhitelist

kripod:
Required for seamless augmentation of React attributes, in order to add the `sx` prop for any element: https://github.com/kripod/glaze/pull/16#issuecomment-616528041

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -184,6 +184,7 @@ sw-toolbox
 swagger-parser
 terser
 three
+treat
 tslint
 ts-toolbelt
 tweetnacl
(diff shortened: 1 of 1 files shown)


kripod:
Thank you!

I’ll try my best to fix `@types/babel-plugin-glaze` as soon as possible. :blush:

sandersn:
It's actually not published yet -- we're on the middle of moving to a monorepo so some previously-automated things are disconnected.

I'll update this thread when things are ready to go.

sandersn:
OK, DefinitelyTyped/DefinitelyTyped#44048 should be ready to update (or close/open if no updates are needed).

I had to copy this commit over to the new monorepo since the npm publisher needs it too.

In the future, the monorepo will publish a separate package for DT's CI, but that's not done yet. @andrewbranch is on it though.

kripod:
Thank you for being so proactive, adding support so fast!

PR #767, opened 2020-04-21
Add dependencies for import types

### LINKED ISSUE #575: import types don't create dependencies
In module-info.ts:imports, ImportType is not a handled SyntaxKind. But they should behave the same as an import declaration.

This caused a problem in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32891 which had to be fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33030 by adding a triple-slash reference, for example.

jablko:
Fixes #575

### DIFF
--- a/src/lib/module-info.ts
+++ b/src/lib/module-info.ts
@@ -265,14 +265,18 @@ function findReferencedFiles(src: ts.SourceFile, packageName: string, subDirecto
  * All strings referenced in `import` statements.
  * Does *not* include <reference> directives.
  */
-function* imports({ statements }: ts.SourceFile | ts.ModuleBlock): Iterable<string> {
-    for (const node of statements) {
+function imports(sourceFile: ts.SourceFile): Iterable<string> {
+    const result: string[] = [];
+    ts.forEachChild(sourceFile, recur);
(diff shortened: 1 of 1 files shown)


sandersn:
Fixes microsoft/DefinitelyTyped-tools#47

src/lib/module-info.ts
```diff
@@ -265,35 +265,43 @@ function findReferencedFiles(src: ts.SourceFile, packageName: string, subDirecto
...
+                    const { literal } = argument as ts.LiteralTypeNode;
+                    if (literal.kind === ts.SyntaxKind.StringLiteral) {
+                        result.push(literal.text);
+                    }
                 }
                 break;
             }
 
             default:
+                ts.forEachChild(node, recur);
```
sandersn, line 304: Does this also fix microsoft/DefinitelyTyped-tools#44? Probably not, but the additional tree walking surely makes some difference.

sandersn (changes requested):
Looks good; same comment as others: needs a test and port to microsoft/DefinitelyTyped-tools

PR #769, opened 2020-04-22
Whitelist `@jest/jest-diff` and `@jest/pretty-format`

douglascayers:
I think these changes are necessary to support https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44033.

The DefinitelyTyped [jest-environment-puppeteer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest-environment-puppeteer/index.d.ts) package has transitive dependencies on `@jest` and the DefinitelyTyped CI is [failing](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44033#issuecomment-616241339) because it can't resolve the external dependencies to `@jest/jest-diff` ([source](https://github.com/facebook/jest/tree/master/packages/jest-diff)) and `@jest/pretty-format` ([source](https://github.com/facebook/jest/tree/master/packages/pretty-format)).

Thanks!

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -14,6 +14,8 @@
 @hapi/wreck
 @jest/environment
 @jest/fake-timers
+@jest/jest-diff
+@jest/pretty-format
 @jest/transform
 @jest/types
 @loadable/component


dependenciesWhitelist.txt
```diff
@@ -14,6 +14,8 @@
 @hapi/wreck
 @jest/environment
 @jest/fake-timers
+@jest/jest-diff
+@jest/pretty-format
```
sandersn, line 18: These aren't on npm, but `jest-diff` and `pretty-format` are. Is there some reason these are not published in the `@jest` namespace?

* https://www.npmjs.com/package/@jest/jest-diff - not found
* https://www.npmjs.com/package/@jest/pretty-format - not found
* https://www.npmjs.com/package/jest-diff - found (and its github repo points to the jest monorepo)
* https://www.npmjs.com/package/pretty-format - found (and its github repo points to the jest monorepo)

douglascayers: @sandersn I have no idea since I don't work on the jest codebase. I'm trying to submit changes over on DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44033) to update typings for `jest-environment-puppeteer` and that [CI process is failing](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44033#issuecomment-616241339) because it can't resolve the `jest-diff` and `pretty-format` modules -- which aren't explicitly referenced in the package I'm trying to update, which is more baffling.

I'm trying to figure out what configuration is needed to get the DefinitelyTyped build process to succeed (I've left many comments over there on the PR with my research and attempts so far).

My latest research is documented at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44033#issuecomment-617504889 if you have any ideas what the DefinitelyTyped error messages mean.

Thanks

douglascayers:
Closing for now as @jablko discovered a [different route](https://github.com/facebook/jest/pull/9875) that may resolve our build errors.

sandersn:
That seems like the right fix. In case you need to revive this PR in the future, I think that it just needs to add `jest-diff` and `pretty-format` instead of `@jest/jest-diff` and `@jest/pretty-format`.

PR #771, opened 2020-04-22
Deduplicate dependencies

jablko:
What do you think about changing the `dependencies` data from an array to an object, to prevent duplicates, as in e.g. https://www.npmjs.com/package/@types/jsdom#additional-details

### Before

```JSON
            "dependencies": [
                {
                    "name": "parse5",
                    "version": "*"
                },
                {
                    "name": "tough-cookie",
                    "version": "*"
                },
                {
                    "name": "node",
                    "version": "*"
                },
                {
                    "name": "parse5",
                    "version": "*"
                },
                {
                    "name": "tough-cookie",
                    "version": "*"
                },
                {
                    "name": "node",
                    "version": "*"
                },
                {
                    "name": "parse5",
                    "version": "*"
                },
                {
                    "name": "tough-cookie",
                    "version": "*"
                },
                {
                    "name": "node",
                    "version": "*"
                },
                {
                    "name": "parse5",
                    "version": "*"
                },
                {
                    "name": "tough-cookie",
                    "version": "*"
                },
                {
                    "name": "node",
                    "version": "*"
                },
                {
                    "name": "parse5",
                    "version": "*"
                },
                {
                    "name": "tough-cookie",
                    "version": "*"
                },
                {
                    "name": "node",
                    "version": "*"
                }
            ],
```

### After

```JSON
            "dependencies": {
                "parse5": "*",
                "tough-cookie": "*",
                "node": "*"
            },
```

### DIFF
--- a/src/lib/definition-parser.ts
+++ b/src/lib/definition-parser.ts
@@ -8,9 +8,9 @@ import {
 
 import { allReferencedFiles, createSourceFile, getModuleInfo, getTestDependencies } from "./module-info";
 import {
+    DependencyVersion,
     formatTypingVersion,
     getLicenseFromPackageJson,
-    PackageId,
     PackageJsonDependency,
     PathMapping,
     TypingsDataRaw,
@@ -170,7 +170,7 @@ function combineDataForAllTypesVersions(
         files,
         license,
         // TODO: Explicit type arguments shouldn't be necessary. https://github.com/Microsoft/TypeScript/issues/27507
-        dependencies: getAllUniqueValues<"dependencies", PackageId>(allTypesVersions, "dependencies"),
+        dependencies: Object.assign({}, ...allTypesVersions.map(({ dependencies }) => dependencies)),
         testDependencies: getAllUniqueValues<"testDependencies", string>(allTypesVersions, "testDependencies"),
         pathMappings: getAllUniqueValues<"pathMappings", PathMapping>(allTypesVersions, "pathMappings"),
         packageJsonDependencies,
@@ -187,7 +187,7 @@ function getAllUniqueValues<K extends string, T>(records: ReadonlyArray<Record<K
 interface TypingDataFromIndividualTypeScriptVersion {
     /** Undefined for root (which uses `// TypeScript Version: ` comment instead) */
     readonly typescriptVersion: TypeScriptVersion | undefined;
-    readonly dependencies: ReadonlyArray<PackageId>;
+    readonly dependencies: { readonly [name: string]: DependencyVersion };
     readonly testDependencies: ReadonlyArray<string>;
     readonly pathMappings: ReadonlyArray<PathMapping>;
     readonly declFiles: ReadonlyArray<string>;
@@ -322,7 +322,10 @@ interface TsConfig {
 }
 
 /** In addition to dependencies found in source code, also get dependencies from tsconfig. */
-interface DependenciesAndPathMappings { readonly dependencies: ReadonlyArray<PackageId>; readonly pathMappings: ReadonlyArray<PathMapping>; }
+interface DependenciesAndPathMappings {
+    readonly dependencies: { readonly [name: string]: DependencyVersion };
+    readonly pathMappings: ReadonlyArray<PathMapping>;
+}
 function calculateDependencies(
     packageName: string,
     tsconfig: TsConfig,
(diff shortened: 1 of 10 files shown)


sandersn (commented):
This seems like a good idea. 3 things:

1. Are there are any pros/cons from the array vs object format? I can't even find documentation of the array format.
3. What about existing packages? I guess they don't need to be republished.
3. We're in the process of moving this repo to a monorepo. Unfortunately, it's held up in the legal process of open-sourcing the new repo. I *think* this is reasonably easy to port, so it might be worth taking before the new repo is available.

Edit: Four things.

4. How widespread is this problem right now?

sandersn:
Actually, poking around unpkg, it looks like npm converts arrays to objects upon publication. So this would only improve the quality of the temporary files right before publication.

jablko:
> 3. it might be worth taking before the new repo is available.

Sounds good :+1:

### Pros of the array format/cons of the object format

I'm not aware of any: As you say, the current format gets flattened into an object anyway, and npm allows at most one semver per dependency, so you can't represent anything legal in the array that you can't with an object.

### Pros of the object format

The main one is that it makes the code behave as written: `combineDataForAllTypesVersions()` calls `getAllUniqueValues()` on `dependencies`, but `getAllUniqueValues()` only does shallow (`===`) equality: https://github.com/microsoft/types-publisher/blob/58735580e2648b9e273d4a7dee76a88394063f16/src/util/util.ts#L91-L93

Instead of fixing/removing `getAllUniqueValues()` on the current format, substituting an object makes it behave as written (unique dependencies) and fixes a consequent trivial bug in the `README.md`/[package details](https://www.npmjs.com/package/@types/jsdom#additional-details). Agreed, existing packages don't need to be republished for this.

As long as we're comparing pros/cons, we don't need to use `Array.some()` [for object lookup](https://github.com/microsoft/types-publisher/pull/771/files#diff-2c6c4bd69c5bde75e2d62e313ce91d28L391): `dependencies.some(d => d.name === dependency)` -> `dependencies[dependency]` FWIW.

### How widespread is this problem right now?

It's limited to the following packages:
```Shell
$ jq 'map_values(select(.[].dependencies | unique != sort)) | keys' data/definitions.json 
[
  "eslint",
  "google-cloud__tasks",
  "jsdom",
  "ng-flow",
  "openpgp",
  "parse",
  "react-event-listener",
  "redis",
  "sinon",
  "styled-components"
]
```

In other words all packages with typesVersions and dependencies:
```Shell
$ jq 'map_values(select(any(.[]; .typesVersions != [] and .dependencies != []))) | keys' data/definitions.json 
[
  "eslint",
  "google-cloud__tasks",
  "jsdom",
  "ng-flow",
  "openpgp",
  "parse",
  "react-event-listener",
  "redis",
  "sinon",
  "styled-components"
]
```

src/lib/definition-parser.ts
```diff
@@ -170,7 +170,7 @@ function combineDataForAllTypesVersions(
         files,
         license,
         // TODO: Explicit type arguments shouldn't be necessary. https://github.com/Microsoft/TypeScript/issues/27507
```
sandersn, line 172: should remove this TODO now

src/lib/definition-parser.ts
```diff
@@ -187,7 +187,7 @@ function getAllUniqueValues<K extends string, T>(records: ReadonlyArray<Record<K
 interface TypingDataFromIndividualTypeScriptVersion {
     /** Undefined for root (which uses `// TypeScript Version: ` comment instead) */
     readonly typescriptVersion: TypeScriptVersion | undefined;
-    readonly dependencies: ReadonlyArray<PackageId>;
+    readonly dependencies: { readonly [name: string]: DependencyVersion };
```
sandersn, line 190: Is PackageId used elsewhere now? If it's not, then it would be worthwhile to change its definition to be closer to this new one.

sandersn: (esp since `PackageId[] -> { readonly [name: string]: DependencyVersion }` in all cases, I think)

sandersn (approved):
Couple of minor comments -- I'll port this over to microsoft/DefinitelyTyped-tools since it's basically ready to go.

sandersn:
OK, microsoft/DefinitelyTyped-tools#57 is merged! Thanks @jablko!

PR #774, opened 2020-04-28
Reference the same files as the compiler

jablko:
types-publisher always resolves non-relative imports relative to the current older-version directory, e.g. `import { createRenderer } from 'react-test-renderer/shallow';` -> `types/react-test-renderer/v15/shallow/index.d.ts`. This implies a `"react-test-renderer/*": ["react-test-renderer/v15/*"]` path mapping, or the compiler will look in the latest-package directory instead (`types/react-test-renderer/shallow/index.d.ts`).

This PR ensures that the path mapping is there. The error is roughly cribbed from https://github.com/microsoft/types-publisher/blob/dd69e3b816be1600d7bec25919c4fe07826573c8/src/lib/definition-parser.ts#L383-L388


sandersn (changes requested):
Same comment as your other fixes. Can you add a test for this, and port it to microsoft/DefinitelyTyped-tools?

PR #776, opened 2020-04-30
Handle more various relative references

jablko:
e.g.
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0d8cc30feb7b4012c98886c8f0112b0a70548cd0/types/dwt/v13/dwt-tests.ts#L1
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0d8cc30feb7b4012c98886c8f0112b0a70548cd0/types/slickgrid/test/slick.autotooltips.ts#L1
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0d8cc30feb7b4012c98886c8f0112b0a70548cd0/types/react/test/experimental.tsx#L1
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0d8cc30feb7b4012c98886c8f0112b0a70548cd0/types/webpack/test/next.ts#L1

### DIFF
--- a/src/lib/module-info.ts
+++ b/src/lib/module-info.ts
@@ -216,8 +216,7 @@ function findReferencedFiles(src: ts.SourceFile, packageName: string, subDirecto
         addReference({ text: ref.fileName, exact: true });
     }
     for (const ref of src.typeReferenceDirectives) {
-        // only <reference types="../packagename/x" /> references are local (or "packagename/x", though in 3.7 that doesn't work in DT).
-        if (ref.fileName.startsWith("../" + packageName + "/")) {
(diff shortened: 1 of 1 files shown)


sandersn (changes requested):
If I recall correctly, this rule is intentional; of the examples you linked to, only the nested reference in slickgrid should be allowed, so probably the change should be to require as many leading `../` as the number of nested directories + 1.

The others should be rewritten to the same form as the slickgrid reference.

(Same comments as other PRs about adding tests and moving to DefinitelyTyped-tools)

PR #778, opened 2020-05-07
Add cash-dom to the dependencies whitelist

chocolateboy:
See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44533


sandersn:

PR #784, opened 2020-05-21
Replace upstreamed dependency (cash-dom) with legacy dependency (@types/cash)

chocolateboy:
This is a continuation of/replacement for #778, implemented as part of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44533: use the legacy cash-dom typings (@types/cash) for its sole dependent (materialize-css) to minimize compatibility issues/breakage.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -24,6 +24,7 @@
 @storybook/addons
 @storybook/react
 @types/base-x
+@types/cash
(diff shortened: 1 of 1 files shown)


sandersn:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44969 passed CI, so I don't think this is needed.

PR #785, opened 2020-05-24
Whitelist @elastic/elasticsearch

pauldraper:
Facilitates https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45016

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -5,6 +5,7 @@
 @babel/template
 @babel/traverse
 @babel/types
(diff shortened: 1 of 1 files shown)


andrewbranch:
👋 @pauldraper thanks, looks good! FYI, this repo has [moved](https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/definitions-parser/dependenciesWhitelist.txt), so I’ll have to replicate this commit in the new location. That should happen later today if all goes well.

PR #786, opened 2020-05-24
feat(mini-html-webpack-plugin): whitelist package

peterblazejewicz:
Part of WebPack ecosystem it now ships its own definitions:
https://github.com/styleguidist/mini-html-webpack-plugin/blob/master/package.json#L6

Thanks!

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -135,6 +135,7 @@ knockout
 magic-string
 mali
 meteor-typings
+mini-html-webpack-plugin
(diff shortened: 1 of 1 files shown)


andrewbranch:
@peterblazejewicz I don’t see any other DefinitelyTyped definitions that depend on this package—I believe it only needs to be added to the whitelist if another package on DefinitelyTyped references it.

peterblazejewicz:
@andrewbranch cool, closing then, thanks!

PR #787, opened 2020-05-25
Add gl-matrix

davidnx:
This is in prep for removing `@types/gl-matrix` from DefinitelyTyped now that [gl-matrix](https://github.com/toji/gl-matrix) ships with its own types.

### DIFF
--- a/dependenciesWhitelist.txt
+++ b/dependenciesWhitelist.txt
@@ -100,6 +100,7 @@ fastify
 final-form
 flatpickr
 form-data
+gl-matrix
 glaze
 globby
 google-auth-library


davidnx:
@RyanCavanaugh @sandersn friendly ping...

andrewbranch:
👋 @davidnx thanks, looks good! FYI, this repo has [moved](https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/definitions-parser/dependenciesWhitelist.txt), so I’ll have to replicate this commit in the new location. That should happen later today if all goes well.

